PR #29, opened 2023-03-03
Rename clientKey to clientId

Cherna:
Rename `clientKey` to `clientId`.

Solves: https://justifi-ai.atlassian.net/browse/FE-46

This change should not really break anything since the work `clientKey` is never actually used externally, just internally.

The call signature is `fn(clientKey, arg2, arg3)` so when calling it, it would always be used as `fn(<clientKey string>, arg2, arg3)`. No explicit use of the keyword `clientKey` or `clientId` (no named arguments as in an object).

Type of Change
--------------

* [ ] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [x] [Potentially] Breaking change (would cause existing functionality to not work as expected)
* [x] This change requires a documentation update
* [x] This change is in scope for PCI
* [ ] This requires approval from UX
* [ ] This requires approval from Marketing

### DIFF
--- a/stencil-library/docs.json
+++ b/stencil-library/docs.json
@@ -1,0 +1,487 @@
+{
+  "timestamp": "2023-03-10T17:22:15",
+  "compiler": {
+    "name": "@stencil/core",
+    "version": "2.20.0",
+    "typescriptVersion": "4.8.4"
+  },
+  "components": [
+    {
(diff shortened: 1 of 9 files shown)


jakemer (approved):
Cross-checked with https://github.com/justifi-tech/payments-js/pull/73 and looks good!

krististepanek (approved):
🍃

PR #32, opened 2023-03-28
Pass component version to paymentsJS

jakemer:
Pass along `componentVersion` as part of the `tokenize` payload, which will be used in the iframed `payments-js` app to include a `x-justifi-component-version` header in tokenize calls. This way when errors are logged for tokenize calls, we know where they are coming from.

Since this change is only being made now, this version of the library and going forward will pass along the component version and old versions will not :/ 

Links
-----
https://justifi-ai.atlassian.net/browse/FE-90

Developer considerations
--------------

Steps for Testing/QA
--------------------
Since this change is only being made now, this version of the library and going forward will pass along the component version and old versions will not :/ 

you will need to inspect the page to find the `iframe` that the components render, and change the `src` to use `http://localhost:3003` so that it is hitting your local `payments-js` with the header changes

### Testing the current components
- Run this branch locally, along with the corresponding `payments-js` [branch](https://github.com/justifi-tech/payments-js/pull/79)
- Go to the billing from example (http://localhost:3333/components/bank-account-form/example.html) and update the `iframe` `src`
- With the developer tools / network tab open, enter valid bank account info and click the "Tokenize" button
- In the developer tools / network tab, there should be a `POST` request to `payment_methods`, and the request headers should include `X-JustiFi-Component-Version: 0.2.1` and `X-JustiFi-PaymentsJS-version: current`
- Go to the card from example (http://localhost:3333/components/card-form/example.html) and repeat the steps 😄

### Testing previous version components
- Download the example files here: [components-examples.zip](https://github.com/justifi-tech/web-component-library/files/11102675/components-examples.zip)
- For previous version components that use the *new* version of `payments-js`, open `previous-version-components.html`
- With the updated/local `iframe` `src`, repeat the steps for `Testing the current components`. There should only be a single header: `X-JustiFi-PaymentsJS-version: current`
- For 'legacy' components that use the *old* version of `payments-js`, open `legacy-card-form.html`
- With the updated/local `iframe` `src`, repeat the steps for `Testing the current components`. There should only be a single header: `X-JustiFi-PaymentsJS-version: legacy`


krististepanek (approved):
🥽

PR #33, opened 2023-03-30
upgrade to latest version of webcomponents

Cherna:
- Change version in react-library to `0.2.0` to match that of `webcomponents`.
- Update `webcomponents` dependency to latest.
- Update package `typescript` and `@types/node` to solve issues.
- Confirmed it built the react-components properly.
- Include changes to lockfile in stencil-library as well, after running install on it.

### DIFF
--- a/react-library/package.json
+++ b/react-library/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justifi/react-components",
   "sideEffects": false,
-  "version": "0.0.14",
+  "version": "0.2.0",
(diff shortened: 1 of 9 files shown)


jakemer (approved):
Looks great 👍 . The react components will include so much more with this version bump!

jakemer:
I was able to check that the react package is building properly with the latest card form using this branch: https://github.com/justifi-tech/web-component-library/compare/main...quick-storybook-for-react-lib

Dev QA ✅

PR #43, opened 2023-04-14
Upgrade sb v7 and add play

Cherna:


stencil-library/.babelrc.json
```diff
@@ -0,0 +1,15 @@
+{
```
jakemer, line 1: Is babel necessary? I don't see how it's being used but I would like to avoid introducing build tools that don't come with Stencil or Storybook out of the box

Cherna: https://storybook.js.org/docs/react/configure/babel

This was a result of migration to the latest version. It automatically created a default babelrc file to keep a babel config somewhere. It was already being used, but the config file was hidden in previous versions.

jakemer: Okay, awesome 👍

jakemer (approved):
Nice, thanks for doing this upgrade!

PR #58, opened 2023-04-24
Fe 131 fix validation strategy in card form

colganmeanor:
### Fix broken validation strategy prop in card form stories

Fixes error when rendering card-form component via storybook. Props controlled via storybook UI should now function correctly, and successfully pass from storybook UI to component render. 

Links
-----

- [Fix 'validationStrategy' in card-form](https://justifi-ai.atlassian.net/browse/FE-131?atlOrigin=eyJpIjoiNTExYjViZDM0ODllNDVhZDhjNGNhMzA5NDViMzIyYmUiLCJwIjoiaiJ9)

Developer considerations
--------------

- [X] Increment the version number for relevant packages
  - [x] This is a new MAJOR version (incompatible API changes were made)
  - [X] This is a new PATCH version (backwards compatible bug fixes)


stencil-library/src/components.d.ts
```diff
@@ -15,7 +15,7 @@ export namespace Components {
         "styleOverrides"?: string;
         "tokenize": (clientId: string, paymentMethodMetadata: any, account?: string) => Promise<any>;
         "validate": () => Promise<any>;
-        "validationStrategy": 'onChange' | 'onBlur' | 'onSubmit' | 'onTouched' | 'all';
+        "validationMode": 'onChange' | 'onBlur' | 'onSubmit' | 'onTouched' | 'all';
```
jakemer, line 18: Since we're changing the public API here, this should actually be considered a "breaking change" (even though this feature was broken before) so the version number should go to `1.0.0`

stencil-library/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file. Dates are d
 
 Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
 
+#### 0.6.2
+> 24 April 2023
+
+- Fix broken prop controller in Storybook UI for card-form component
+- rename variable `validationStrategy` and attr `validation-strategy` to `validationMode` and `validation-mode` respectively
```
jakemer, line 11: Please change `attr` to `prop`, just to be extra clear that it is a stencil prop

jakemer (approved):
Nice work! Thanks for fixing this

PR #62, opened 2023-04-25
improve docs

Cherna:
image


jakemer (approved):
Nice! You may want to bump the patch version in package.json and changelog if you want a new package with these docs changes

jakemer (approved):

PR #65, opened 2023-04-27
add stories for billingform

Cherna:
image
image


stencil-library/docs.json
```diff
@@ -10,7 +10,7 @@
       "filePath": "./src/components/bank-account-form/bank-account-form.tsx",
       "encapsulation": "none",
       "tag": "justifi-bank-account-form",
-      "readme": "# justifi-bank-account-form\n\n\n",
+      "readme": "# tag: <justifi-bank-account-form />\n\n\n",
```
jakemer, line 13: Is it possible to use markdown in this? Just curious. I think IF we can we should do `<justifi-bank-account-form />` (<- like apply the code fencing syntax like this) instead of "tag:". Just a thought, not important 😄

stencil-library/src/components/billing-form/billing-form.scss
```diff
@@ -3,4 +3,52 @@
 :host {
   @include host-base-styles;
   display: block;
-}
\ No newline at end of file
+}
```
jakemer, line 6: For this component the implementer can actually use the css parts directly, because they are exported from billing form. So I think the css variables are only necessary for anything that uses the payment method components (because we need a way to grab the styles via javascript to send through the iframe). We should discuss. I think the CSS parts are a better experience and we should only use the css variables when we have to, but maybe that's confusing

stencil-library/docs.json
Cherna: It could definitely be an option, let me try it out

jakemer (approved):

PR #68, opened 2023-05-03
FE-144: Check for contentWindow before attempting postMessage

jakemer:
There have been a number of occurrences where a `postMessage` is attempted but `iframe.contentWindow` is not defined. This could be a race condition with the iframe being rendered in the DOM, or it could be that the `message` event handler is not being properly cleaned up in all situations, or both. In any case, however, we should handle this more gracefully and check that `iframe.contentWindow` is defined before attempting to call `postMessage` on it

Links
-----

Developer considerations
--------------

Steps for Testing/QA
--------------------
- Pull this branch and run the build from the `stencil-library` directory
- Run `yarn link` from the `stencil-library` (you may already have it linked, that's fine)
- Download this [vue sample app](https://github.com/justifi-tech/web-component-library/files/11401620/vue-sample-app.zip) and unzip it somewhere
- Inside the unzipped vue app directory, run `yarn link "@justifi/webcomponents"`
- Inside the unzipped vue app directory, run `yarn` and then `yarn run dev`
- When the dev server starts, it will say what port it is running on. Open it up in your browser
- The app should consist of the card form component and a button to toggle it (there is a vue template condition around the component)
- With the dev tools open, click the toggle a few times and ensure that there is no error related to `contentWindow` being `null` or `undefined`


stencil-library/src/components/payment-method-form/payment-method-form.tsx
```diff
@@ -60,7 +60,7 @@ export class PaymentMethodForm {
   }
 
   private postMessage(eventType: string, payload?: any) {
-    if (this.iframeElement) {
+    if (this.iframeElement && this.iframeElement.contentWindow) {
```
TiagoFuelber, line 63: Hi @jakemer 
I have a suggestion, instead of using an if we could use optional chaining, which is slightly better to read. WDYT?
`this?.iframeElement?.contentWindow?.postMessage({ eventType: eventType, ...payload }, '*');`

jakemer: @msencenb Yep! It does get transformed in the build pipeline. There is also a `webcomponents.cjs.js` that you can use instead which is common JS

msencenb:
@jakemer To add more context here, I think the reason why I am seeing this in my Vue app is because my component that wraps the justifi-card-form or justifi-payment-form is conditionally rendered with v-if. I believe that means the component element doesn't enter the dom until after the user clicks through a specific step (selecting the billing option they are buying). 

I don't know that it changes the fix at all, but figured I would send over more context as to why this could happen in the wild.

TiagoFuelber (approved):
LGTM

krististepanek (approved):
:octocat:

PR #69, opened 2023-05-04
Add Intro to storybook

Cherna:
image
image


Cherna:
> @Cherna for the React components, are you sure that the global styles aren't needed? Somehow the css variables need to be defined on `:root`. I am realizing that might actually be a problem, I think the React package SHOULD have some css files in it doesn't look like it does

Yeah, I understand it might not be the case right now, it was more of a "the way I think it should be".

jakemer (approved):

PR #72, opened 2023-05-08
add styling for radio buttons

Cherna:
Screenshots
-----

image
image

Links
-----

https://justifi-ai.atlassian.net/browse/FE-142

Developer considerations
--------------


stencil-library/src/components/payment-form/payment-form.scss
```diff
@@ -54,4 +54,10 @@ justifi-payment-method-form {
     border-color: var(--jfi-form-control-border-color-error);
     box-shadow: var(--jfi-form-control-box-shadow-error-focus);
   }
-}
\ No newline at end of file
+}
+
+.submit-button {
+  color: var(--jfi-form-control-submit-button-color);
```
jakemer, line 60: `form-control` is meant to mean input or select, but I don't think it is needed as part of the submit button variable name

stencil-library/src/styles/root.scss
```diff
@@ -30,6 +30,16 @@
   --jfi-form-control-margin: 0;
   --jfi-form-control-padding: #{$input-padding-y $input-padding-x};
 
+  // form radio group
+  --jfi-form-control-radio-button-color: #{$primary};
```
jakemer, line 34: Could we just have one set of new variables called `--jfi-button-{property}`?  To bootstrap, both are buttons and we should be able to apply the variables to both the button group and submit button

stencil-library/src/components/payment-form/payment-form.tsx
```diff
@@ -103,7 +103,7 @@ export class PaymentForm {
               onClick={(event) => this.submit(event)}
               disabled={!this.submitButtonEnabled}
               type="submit"
-              class="btn btn-primary">
+              class="btn btn-primary submit-button">
```
jakemer, line 106: Do we need a new class or can we just override bootstrap's `btn` class?

Cherna: I'm ok with overriding it if it won't apply anywhere outside of the `payment-form.tsx` component file.

stencil-library/src/styles/root.scss
Cherna: The buttons within groups and outside of them need different styling, since their default state (i.e. when not "checked" in the case of button groups) don't match.

Cherna: What we could do is add some variables that are used in other variables to easily override many at the same time, like:
`--jfi-primary: {#primary}`. Then that can be used in many places at the same time so you can override them separately, but also just change the primary color and change them all at once.

stencil-library/src/components/payment-form/payment-form.tsx
Cherna: Tried it and turns out we run into a rule specificity issue. Using a new class both solves that and keeps it better scoped.

jakemer (approved):

PR #78, opened 2023-05-11
update config to allow JSX and change readme to custom template without StoriesOf

Cherna:
Developer considerations
--------------


stencil-library/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file. Dates are d
 
 Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
 
+#### 3.2.5
```
jakemer, line 7: Since this is a Storybook change only, it should not be a new version (a dev would not get anything by upgrading from 3.1.5 to 3.2.5 in their application)

stencil-library/.storybook/preview.jsx
```diff
@@ -13,6 +15,15 @@ export const parameters = {
   actions: { argTypesRegex: "^on.*" },
   docs: {
     extractArgTypes: extractArgTypesFactory({ dashCase: true }),
+    page: () => (
```
jakemer, line 18: So does this mean that our stories can also use jsx? Because that would be awesome

jakemer (approved):

PR #82, opened 2023-05-12
update story docs to more flexible approach and update examples

Cherna:
- Remove the auto-generated readme files and replaced them with more extensible and flexible `.stories.mdx` files.
- Generate JSON docs and create a function to filter by `<tag>`, not used yet but important to get any components docs from within any docs file.
- Update examples for each component, mainly by removing outdated info, like `style-overrides` or old definition of variables to style.

Developer considerations
--------------


stencil-library/src/components/card-form/example.js
```diff
@@ -1,53 +1,6 @@
-<!DOCTYPE html>
-<html dir="ltr" lang="en">
-
-<head>
-  <meta charset="utf-8" />
-  <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0" />
-  <title>card-form component</title>
-
-  <script type="module" src="../../build/webcomponents.esm.js"></script>
```
jakemer, line 9: We used to use these files locally for development, but now that they are just code samples we should update the src to be the cdn link pointing to `latest`

Cherna: This code is being deleted anyways, since for example usage, it's not relevant how you decided to include the files, that's described in the Introduction page.

jakemer (approved):
Lots of great improvements!

PR #105, opened 2023-06-14
FE-173 add unit and e2e tests to payment-form

TiagoFuelber:
* Added unit and end-to-end tests to payment-form component

Links
-----

Developer considerations
--------------

Steps for Testing/QA
--------------------
No QA needed

### DIFF
--- a/stencil-library/src/components/payment-form/test/payment-form.e2e.tsx
+++ b/stencil-library/src/components/payment-form/test/payment-form.e2e.tsx
@@ -1,0 +1,33 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('justifi-payment-form', () => {
+  it('should correctly receive and set the props', async () => {
+    // Create new testing page and add our component to it
+    const page = await newE2EPage();
+    await page.setContent(
+      `
+        <justifi-payment-form
+          bank-account="true"
+          card="true"
+          email="test@example.com"
+          iframe-origin="https://iframe.example.com"
+          client-id="testId"
+          account-id="accountId"
+          submit-button-text="Pay Now"
+        />
+      `,
+    );
+
+    // Find our component on the page
+    const element = await page.find('justifi-payment-form');
+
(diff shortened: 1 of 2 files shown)


stencil-library/src/components/payment-form/test/payment-form.e2e.tsx
```diff
@@ -0,0 +1,33 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('justifi-payment-form', () => {
+  it('should correctly receive and set the props', async () => {
```
jakemer, line 4: For the e2e tests I think it probably makes more sense to test how the props affect the UI vs checking that it has the expected props. For example, `submit-button-text="Pay Now"`, should actually cause the button to have that text. For the unit tests maybe we can have one that is something like this:
```
const paymentForm = new PaymentForm();
expect(paymentForm).toHaveProperty('submitButtonText', undefined);
```
Just to make sure that the properties exist, so that if they are ever removed from the component we are aware of the breaking change?

stencil-library/src/components/payment-form/test/payment-form.spec.tsx
```diff
@@ -1,15 +1,207 @@
 import { newSpecPage } from '@stencil/core/testing';
 import { PaymentForm } from '../payment-form';
+import { PaymentMethodTypes } from '../../../api';
+import { BillingFormFields } from '../../billing-form/billing-form-schema';
 
 describe('justifi-payment-form', () => {
```
jakemer, line 6: Very nice coverage in here, this is awesome!

stencil-library/src/components/payment-form/test/payment-form.e2e.tsx
TiagoFuelber: Yes, totally makes sense, thanks for the suggestion.

jakemer (approved):

PR #107, opened 2023-06-20
Form controller based forms

jakemer:
Create a form class to store form values, errors, and a register function for registering inputs with handlers

Links
-----

Developer considerations
--------------

Steps for Testing/QA
--------------------

### DIFF
--- a/stencil-library/src/components/form/form-control-select.tsx
+++ b/stencil-library/src/components/form/form-control-select.tsx
@@ -8,17 +8,18 @@ import { Component, Host, h, Prop } from '@stencil/core';
 export class TextInput {
   @Prop() label: string;
   @Prop() name: any;
-  @Prop() onInput: (e: any) => void;
+  @Prop() onInput: (e: Event) => void;
   @Prop() error: string;
   @Prop() options: { label: string; value: string }[];
 
   render() {
     return (
       <Host exportparts="label,input,input-invalid">
-        <label part="label" class="form-label">
+        <label part="label" class="form-label" htmlFor={this.name}>
           {this.label}
         </label>
         <select
+          id={this.name}
           name={this.name}
           onInput={this.onInput}
           part={`input ${this.error && 'input-invalid'}`}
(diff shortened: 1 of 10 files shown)


stencil-library/src/components/form/form-control-select.tsx
```diff
@@ -0,0 +1,34 @@
+import { Component, Host, h, Prop } from '@stencil/core';
+
+@Component({
+  tag: 'form-control-select',
+  styleUrl: 'form-control-select.scss',
+  shadow: true,
+})
+export class TextInput {
+  @Prop() label: string;
+  @Prop() name: any;
```
TiagoFuelber, line 10: shouldn't it be string type?

stencil-library/src/components/form/form-control-select.tsx
```diff
@@ -0,0 +1,34 @@
...
+  @Prop() onInput: (e: any) => void;
```
TiagoFuelber, line 11: suggestion: 
`@Prop() onInput: (e: Event) => void;`

stencil-library/src/components/form/form-control-text.tsx
```diff
@@ -0,0 +1,33 @@
...
+  @Prop() error: string;
+  @Prop() value: string;
+
+  render() {
+    return (
+      <Host exportparts="label,input,input-invalid">
+        <label part="label" class="form-label">
+          {this.label}
+        </label>
+        <input
```
TiagoFuelber, line 21: suggestion/question: 
It would be more accessible if we move the `input` inside the `label`. It would help screen readers know to which field the input is associated.  Would it still work fine with that change?

TiagoFuelber (approved):
Approved with minor comments.

stencil-library/src/components/form/form-control-text.tsx
jakemer: Thats a good catch, I think the label is meant to have a `for` and the input an `id` to let screen readers know the relationship between labels and inputs

TiagoFuelber: Perfect! That will be just fine.

TiagoFuelber (approved):

PR #109, opened 2023-06-21
add spinner and missing var

Cherna:
QA plan:
- Submit a payment on a hosted checkout session. (Sorry you have to run it locally Kristi 😞 )

Expect:
- Payment form works the same as it was, except there's a spinner while submitting.
- When the Payment form is styled, the loading state is styled properly and not change unexpectedly.

Changes:
- Add separate loading state from enabled state.
- Add missing variables for loading state.
- Add spinner when loading.

image
image


TiagoFuelber (approved):

krististepanek (approved):
🐘

PR #111, opened 2023-06-21
Rename files and classes to be consistent with other forms

jakemer:
- Rename files to use `-form` naming convention
- Rename other occurrences of `businessInfo` to `business`
- Move `business-address-form` component so that it is a subdirectory of `business-form`

Steps for Testing/QA
--------------------
- Pull this branch
- from the `stencil-library` folder, run `yarn start`
- Go to `http://localhost:3334/components/business-form/example.html`
- If all is well, the business form component should render and validation should work when clicking submit 😄


TiagoFuelber (approved):
Good one!

Cherna (approved):

PR #118, opened 2023-06-27
FE-234 ignore auto generated files

TiagoFuelber:
* Deleted auto-generated files
* Added those files to .gitignore

Links
-----

Developer considerations
--------------

Steps for Testing/QA
--------------------


Cherna (approved):

stencil-library/.gitignore
```diff
@@ -1,6 +1,12 @@
+# Auto-generated
 dist/
 www/
 loader/
+docs.json
+build/
+docs.d.ts
+src/components.d.ts
+src/components/*/readme.md
```
jakemer, line 9: Do we want to be able to see these in github? I'm not sure how useful it is, but it may be nice to see in github vs going to storybook instead

TiagoFuelber: Yes, good point. Also this one doesn't change a lot, like the docs.json for example.

jakemer (approved):

PR #140, opened 2023-07-17
Handling complex forms

jakemer:
Using a [FormController](https://github.com/justifi-tech/web-component-library/pull/140/files?file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx&show-deleted-files=true&show-viewed-files=true#diff-dd0995223e6369211415bc14a805ff86adeb818c4f8d39b19fc09252a7d00006) that provides observables, we can break up large forms into multiple components that share the same state.

In this PR we have a working form with sub-components:

```
- BusinessForm
   - BusinessGenericInfo
   - BusinessRepresentative
     - BusinessAddress (generic component)
```

These form components use an instance of a `FormController`, which returns different pieces of form state as observables so that they can be subscribed to.

Note:
`BusinessAddress` is a generic component, so it doesn't subscribe directly to the form controller. This is because the component, due to being generic, does not have the context it needs in order to properly apply it's values. So instead it's provided with a `onFormUpdate` handler function which is called when the `BusinessAddress` internal values change. This way the the parent component (`BusinessRepresentative` in the case of this PR) can choose how to apply the values supplied by `BusinessAddress` to the form state.

Links
-----

Developer considerations
--------------

Steps for Testing/QA
--------------------


stencil-library/src/components/form/form.ts
```diff
@@ -0,0 +1,67 @@
...
+  public async validateAndSubmit(submitHandler: () => void): Promise<void> {
+    const isValid = await this.validate();
+    if (isValid) {
+      submitHandler();
+    }
+  }
+
+  public setValues(values): void {
+    this._values = { ...this._values, ...values };
+    this.values.next(this._values);
```
TiagoFuelber, line 60: interesting

jakemer: Yep! This is how you push data into the observable so that it's updated everywhere

Cherna (approved):

TiagoFuelber (approved):
Looks good to me. Nice!

PR #144, opened 2023-07-20
Add slot and story

Cherna:
image

Developer considerations
--------------

  - [x] This is a new MINOR version (functionality was added in a backwards compatible manner)


stencil-library/src/components/payment-form/payment-form.stories.ts
```diff
@@ -148,3 +151,19 @@ Styled.args = {
   --jfi-radio-button-group-width: 100%;
   `,
 };
+
+export const Slotted = Template.bind({});
```
jakemer, line 155: I don't think we should add a story until we actually have a insurance component to show here

stencil-library/src/components/payment-form/payment-form.tsx
```diff
@@ -103,6 +103,7 @@ export class PaymentForm {
               }}
             />
           </div>
+          <slot name='insurance' />
```
jakemer, line 106: Do you see any css issues related to inherited styles since this is a direct child of an element with the `row gy-3` classes? I am wondering if it should be wrapped in a grid element, however if there is nothing in the slot we don't want a random space. Not sure what the best way to handle this is off the top of my head

jakemer (changes requested):
This will also need a changelog entry and a minor version bump

stencil-library/src/components/payment-form/payment-form.tsx
Cherna: Interesting, I hadn't considered that to be honest.

Now that I look at how it works, it seems the slot doesn't inherit anything except from :host
image

Cherna: This can be a problem, currently it inherits the following:
image
I don't think that will mess anything up, but we can definitely try it out and iterate if it's an issue.

jakemer (approved):

jakemer (approved):

PR #150, opened 2023-07-31
Fe 248 generic table shared components

Cherna:
Developer considerations
--------------

### DIFF
--- a/stencil-library/src/components/table/table.tsx
+++ b/stencil-library/src/components/table/table.tsx
@@ -1,0 +1,138 @@
+import { Component, Host, h, Prop } from '@stencil/core';
+import { PagingInfo, pagingDefaults } from '../table/table-utils';
+
+interface ExtendedPagingInfo extends PagingInfo {
+  onPrev: () => void;
+  onNext: () => void;
+}
+
+const ExtendedPagingDefaults: ExtendedPagingInfo = {
+  ...pagingDefaults,
+  onPrev: () => {},
+  onNext: () => {},
+}
+
+export interface TableProps {
+  loading: boolean;
+  errorMessage: string;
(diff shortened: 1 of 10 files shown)


stencil-library/src/components/table/table.tsx
```diff
@@ -0,0 +1,138 @@
...
+        <div part="loading-state-spinner" class="spinner-border" role="status">
+          <span class="visually-hidden">Loading...</span>
+        </div>
+      </td>
+    </tr>
+  );
+
+  paginationBar = () => {
+    return (
+      <div class="pagination-bar d-flex justify-content-center gap-3">
```
jakemer, line 66: Are `pagination-bar`, `pagination-btn`, `pagination-prev-btn` etc being used? I don't see the styles anywhere

stencil-library/src/components/table/table.tsx
```diff
@@ -0,0 +1,138 @@
+import { Component, Host, h, Prop } from '@stencil/core';
+import { PagingInfo, pagingDefaults } from '../table/table-utils';
+
+interface ExtendedPagingInfo extends PagingInfo {
+  onPrev: () => void;
+  onNext: () => void;
+}
+
+const ExtendedPagingDefaults = {
```
jakemer, line 9: Should this be `const ExtendedPagingDefaults: ExtendedPagingInfo =`?

Cherna: Yes! Good catch.

stencil-library/src/components/table/table.tsx
Cherna: Nope, can be removed safely.

jakemer (approved):

PR #151, opened 2023-08-03
feat: add proceeds list component

Cherna:
Developer considerations
--------------

image

### DIFF
--- a/stencil-library/src/components/proceeds-list/proceeds-list.tsx
+++ b/stencil-library/src/components/proceeds-list/proceeds-list.tsx
@@ -1,0 +1,145 @@
+import { Component, Host, h, Prop, State, Watch } from '@stencil/core';
+import { Api, IApiResponseCollection } from '../../api';
+import { formatDate } from '../../utils/utils';
+import { PagingInfo, pagingDefaults } from '../table/table-utils';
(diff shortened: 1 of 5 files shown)


stencil-library/src/components/proceeds-list/proceeds-list.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  async fetchData(direction?: string): Promise<void> {
+    if (!this.accountId || !this.authToken) {
+      this.errorMessage = "Can not fetch any data without an AccountID and an AuthToken";
+      this.loading = false;
+      return;
+    }
+    this.loading = true;
+    const limit = this.paging.amount;
+    const cursor = `${
```
jakemer, line 77: I think this could be made to work more like our dashboard app - it just sends the pagination params along to `Api.ts` and is converted to url query params there. Would you please see if that would be an easy change?

jakemer (changes requested):

jakemer (approved):
Signing off on this with the ask that we handle pagination by letting the components manage a params state which is passed along with api calls, where Api.ts can form the params and endpoint into a proper request url like it does in our other apps

PR #164, opened 2023-08-15
feat: add titles to table headers, fix small issue with env

Cherna:
Developer considerations
--------------


stencil-library/src/components/table/table.tsx
```diff
@@ -93,8 +93,8 @@ export class Table {
             <tr class="table-light" part='table-head-row'>
               {
                 this.columnData?.map((column) =>
-                  <th part="table-head-cell" scope="col">
-                    {column}
+                  <th part="table-head-cell" scope="col" title={Array.isArray(column) ? column[1] : ''}>
```
TiagoFuelber, line 96: As a suggestion, not necessary, but we could destruct the array like: 
const [title, content] = column;

Just to have a variable name for column[0] and column[1]

PR #168, opened 2023-08-18
fix(angular-library): update package.json

TiagoFuelber:
fix(angular-library): add main and types fields

we needs this fields in order to import correctly the components using `@justifi/angular-components`

Links
-----

Developer considerations
--------------

Steps for Testing/QA
--------------------

### DIFF
--- a/angular-library/package.json
+++ b/angular-library/package.json
@@ -1,6 +1,8 @@
 {
   "name": "@justifi/angular-components",
   "version": "4.4.0",
+  "main": "dist/component-library/fesm2022/component-library.mjs",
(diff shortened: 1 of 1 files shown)


angular-library/package.json
```diff
@@ -1,6 +1,8 @@
 {
   "name": "@justifi/angular-components",
   "version": "4.4.0",
```
jakemer, line 3: Should the version get bumped to a patch version?

TiagoFuelber: Yes, indeed.

PR #169, opened 2023-08-22
FE-243 vue components

TiagoFuelber:
Adds the Vue Output target to generate Vue components

- the components will e generated when run `npm run build` in the root of the project

Links
-----

Developer considerations
--------------

Steps for Testing/QA
--------------------


vue-library/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "outDir": "./dist",
     "lib": ["dom", "es2020"],
-    "module": "es2015",
+    "module": "commonjs",
```
jakemer, line 6: We might want to come back to this later and try to get this to work as es2015 again as it is old but still more modern than common js

jakemer (dismissed):

vue-library/tsconfig.json
TiagoFuelber: actually, I figured it out this change is not necessary. I am reverting it

jakemer (approved):

PR #177, opened 2023-09-06
feat: add Detail component, add PayoutDetails, fix dev env

Cherna:
Developer considerations
--------------

image


stencil-library/src/components/details/details.tsx
```diff
@@ -0,0 +1,97 @@
...
+        detail-method-title,detail-method-data'
+      >
+        {
+        this.errorMessage ?
+          ErrorState(this.errorMessage)
+        :
+          <main class="p-2">
+            <div part='detail-head' class="p-2">
+              <div class="d-flex flex-row align-items-center gap-2 mb-2">
+                <h1 class="m-0 text-uppercase" part='detail-title'>{this.entity.title}</h1>
```
jakemer, line 45: I think this probably should not be an h1, but if you need the text to be the size of an h1 maybe there is a bootstrap utility class for font size? I also think it should not be `text-uppercase`, but instead the text passed should be uppercase when appropriate (like the USD in your screenshot)

stencil-library/src/components/details/details.tsx
```diff
@@ -0,0 +1,97 @@
+import { Component, Host, h, Prop } from '@stencil/core';
+import { ErrorState } from './utils';
+
+interface IEntityProps {
+  title: string;
+  entityHeadInfo: {
+    title: string,
+    value: any
+  }[];
+  entitySections: {
```
jakemer, line 10: Instead of passing all of this config I think we should leverage named slots and subcomponents such as `DetailSectionTitle` etc

stencil-library/src/components/details/details.tsx
```diff
@@ -0,0 +1,97 @@
...
+                      <span part="detail-head-info-item-data" class="pt-2 ps-2 pe-2">{item.value}</span>
+                    </div>
+                  )
+                }
+              </div>
+            </div>
+            {
+              this.entity.entitySections.map(section =>
+                <div part="detail-details" class="p-2 mt-4">
+                  <h2 part="detail-details-title" class="fw-bold fs-1">{section.sectionTitle}</h2>
```
jakemer, line 64: I believe in the mocks these are `h5`s

jakemer (changes requested):

stencil-library/src/components/details/details.tsx
Cherna: About it being an `<h1>`, it's in the design, but I'm happy to change it. Do you think it might be an issue for screen readers?

About the uppercase, I agree, I can uppercase the text when passing it. I'll change that.

stencil-library/src/components/details/details.tsx
Cherna: I may be misreading the Figma, but it says H2 there:
image

stencil-library/src/components/details/details.tsx
Cherna: Will take a look and see if I find a better approach.

stencil-library/src/components/details/details.tsx
Cherna: Oh, I think I may be looking at an older version.

jakemer (approved):

PR #183, opened 2023-09-20
FE-348 improve inputs

TiagoFuelber:
* Added masked inputs for money, date and phone

Links
-----

Developer considerations
--------------

Steps for Testing/QA
--------------------

### DIFF
--- a/stencil-library/src/components/form/form-control-monetary.tsx
+++ b/stencil-library/src/components/form/form-control-monetary.tsx
@@ -1,0 +1,91 @@
+import {
+  Component,
+  Host,
+  h,
+  Prop,
+  Event,
+  EventEmitter,
+  Element,
+  Watch,
+} from '@stencil/core';
+import IMask, { InputMask } from 'imask';
+
+@Component({
+  tag: 'form-control-monetary',
+  styleUrl: 'form-control-number.scss',
+  shadow: true,
+})
+export class MonetaryInput {
+  @Element() el: HTMLElement;
+
+  @Prop() label: string;
+  @Prop() name: any;
+  @Prop() error: string;
+  @Prop() defaultValue: string;
+  @Prop() inputHandler: (name: string, value: string) => void;
+
+  private imask: InputMask<any> | null = null;
+
+  @Event() formControlInput: EventEmitter<any>;
+  @Event() formControlBlur: EventEmitter<any>;
+
+  @Watch('defaultValue')
+  handleDefaultValueChange(newValue: string) {
+    this.updateInput(newValue);
+  }
+
+  componentDidLoad() {
+    const inputElement = this.el.shadowRoot.querySelector('input');
+    if (inputElement) {
+      this.imask = IMask(inputElement, {
+        mask: Number,
+        scale: 2,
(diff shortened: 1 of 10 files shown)


stencil-library/src/components/form/form-control-monetary.tsx
```diff
@@ -0,0 +1,89 @@
...
+  @Event() formControlInput: EventEmitter<any>;
+  @Event() formControlBlur: EventEmitter<any>;
+
+  @Watch('defaultValue')
+  handleDefaultValueChange(newValue: string) {
+    this.updateInput(newValue);
+  }
+
+  componentDidLoad() {
+    const inputElement = this.el.shadowRoot.querySelector('input');
```
jakemer, line 38: Can we use a `ref` instead? I believe it's supposed to be more reliable: https://stenciljs.com/docs/templating-jsx#getting-a-reference-to-a-dom-element

stencil-library/src/components/form/form-control-number.scss
```diff
@@ -17,3 +17,17 @@ input[type='number']::-webkit-outer-spin-button {
 input[type='number'] {
   -moz-appearance: textfield;
 }
+
+input[type='number'][disabled] {
```
jakemer, line 21: Instead of creating styles, can we use the bootstrap default styles that get applied to `disabled` form controls? https://getbootstrap.com/docs/5.3/forms/select/#disabled

It looks like you can also use the `disabled` attribute on a `fieldset` to disable the whole fieldset as shown here: https://getbootstrap.com/docs/5.3/forms/overview/#disabled-forms

jakemer (changes requested):

stencil-library/src/components/form/form-control-monetary.tsx
TiagoFuelber: Oh sure, I forgot about that!

stencil-library/src/components/form/form-control-number.scss
TiagoFuelber: Nice!

stencil-library/src/components/form/form-control-number-masked.tsx
```diff
@@ -37,9 +36,8 @@ export class NumberInputMasked {
   }
 
   componentDidLoad() {
-    const inputElement = this.el.shadowRoot.querySelector('input');
-    if (inputElement && this.mask) {
-      this.imask = IMask(inputElement, {
+    if (this.textInput && this.mask) {
```
jakemer, line 39: Does this condition need the `&& this.mask`? I noticed that the other form controls don't check for it (or maybe those need it?)

TiagoFuelber: Actually, it's just an extra check in case the component don't receive any mask so the logic won't break. The other components don't have the mask prop.

jakemer: Ahh, okay. Thanks for the clarification!

jakemer (approved):

PR #190, opened 2023-10-10
feat: Business Details

TiagoFuelber:
* Initial HTML structure
* Fetches data
* Handles loading state

image

Links
-----

Developer considerations
--------------

  - [x] This is a new MINOR version (functionality was added in a backwards compatible manner)

Steps for Testing/QA
--------------------

* Go to http://localhost:6006/?path=/story/components-businessdetails--basic
* Add the `auth-token` and a `business-id` props
* If wrong token or ID is passed an error must appear
* If the right props are passed the right data must be displayed


stencil-library/src/api/Business.ts
```diff
@@ -13,96 +13,96 @@ export enum BusinessStructure {
...
-  platform_account_id: string,
-  line1: string,
-  line2: string,
-  postal_code: string,
-  city: string,
-  state: string,
-  country: string,
-  created_at: string
-  updated_at: string
+  id: string;
```
jakemer, line 27: Thanks for fixing these!

stencil-library/src/components/business-details/business-details.tsx
```diff
@@ -0,0 +1,95 @@
...
+      return LoadingState;
+    }
+
+    if (this.renderState === RENDER_STATES.ERROR) {
+      return ErrorState(this.errorMessage);
+    }
+
+    return (
+      <Host>
+        <business-details-section business={this.business} />
```
jakemer, line 89: Can we be consistent with our business form sections? The mocks may only be 3 sections but I think we should break it up in the same way that we did our business form:
```
business-generic-info
legal-address-form
business-representative
business-owners
additional-questions
```
which would be:
```
business-generic-info-details
legal-address-form-details
business-representative-details
business-owners-details
additional-questions-details
```

@nhalm are you okay with that? I think it strays from the mocks but it gives us more modularity and the possibility to "hot swap" sections in the case we would want to offer a UI with read only and read/write sections, or if a customer wanted to create their own. Generally probably good to keep them aligned.

TiagoFuelber: I think that makes perfect sense. I'll make the changes.

colganmeanor (approved):
Looks good Tiago - great job!

stencil-library/src/components/business-details/shared-business-details.scss
```diff
@@ -0,0 +1,60 @@
...
+}
+
+::part(flex-1) {
+  flex: 1;
+}
+
+// Loading and Error States
+.loading-state,
+.error-state {
+  font-size: 1.2rem;
```
jakemer, line 57: Since these are shared components, the loading and error states, should these styles be directly applied to those so that everything is consistent? Also there are bootstrap utility classes for all three of these properties that we can use:
https://getbootstrap.com/docs/5.3/utilities/text/
https://getbootstrap.com/docs/5.3/utilities/spacing/

stencil-library/src/components/business-details/shared-business-details.scss
```diff
@@ -0,0 +1,60 @@
...
+
+  &::part(detail-section-title) {
+    font-size: 1.5rem;
+    margin-bottom: 15px;
+  }
+}
+
+// Detail Items
+detail-item {
+  margin-bottom: 10px;
```
jakemer, line 20: Can we use the bootstrap grid or utility classes to accomplish this? https://getbootstrap.com/docs/5.3/utilities/spacing/

jakemer (approved):

Issue #202, opened 2023-10-17
New component: Payment balance transactions table


jakemer:
- Fetch payment balance transactions from `account/${accountId}/payments/${paymentId}/payment_balance_transactions`
- Display the results in a table with `Processed On`, `Type`, `ID`, `Amount`, and `Balance` columns

Look at the screenshot below to see what we are trying to recreate as a web component:

![image](https://github.com/justifi-tech/web-component-library/assets/3867103/48b85ea4-b023-4ddb-afcc-68d6ae656b44)

PR #210, opened 2023-10-23
#202 feat: Add payment-balance-transactions component

TiagoFuelber:
* Add payment-balance-transactions component

image

* Should be the same table displayed on the dashboard on "Payments" -> "Transactions" tab.

image

Links
-----
[Issue #202](#202)

Developer considerations
--------------

  - [X] This is a new MINOR version (functionality was added in a backwards compatible manner)

Steps for Testing/QA
--------------------
* Pull the branch and run the project.
* Go to http://localhost:6006/?path=/story/components-paymentbalancetransactions--basic
* Add the `auth-token`, `payment-id` and `account-id` props
* The table with the transactions should be displayed


jakemer:
Addresses issue #202

angular-library/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ### 4.6.0
 
+> (insert date here upon release)
```
jakemer, line 5: If you want to leave this out you can, and when we do the release we can copy the changelogs over to the other libraries so we only have to do it once. Also it looks like this one is behind, so we will want to fix that when we do the release too!

jakemer (approved):
Nice work! Looks good to me 👍

Issue #196, opened 2023-10-12
Add event usage to all list components documentation


colganmeanor:
- PaymentsList
- PayoutsList
- BusinessList
- ProceedsList

PR #211, opened 2023-10-23
docs: add some documentation for events

TiagoFuelber:
(Most of the changes were just auto-indentation)

* Added documentation comments for List Components events.

Links
-----

[#196](#196)

Developer considerations
--------------

- [x] Update the appropriate documentation. Keep in mind that this mono repository has README files at the repository, project, and component levels. We want developers to be able to find what they need from the main README, so make sure to add links to component documentation if adding something new.

Steps for Testing/QA
--------------------


stencil-library/src/components/business-list/business-list.tsx
```diff
@@ -48,81 +60,88 @@ export class BusinessList {
...
-        return 'For Profit'
+        return 'For Profit';
       case 'government_entity':
-        return 'Government Entity'
+        return 'Government Entity';
     }
-  }
+  };
 
   mapBusinessStructure = (type: string) => {
```
jakemer, line 83: We should probably convert the swtich/cases in this file to objects at some point... not necessarily in this PR since you did not create them, but something to note I guess

jakemer (approved):
Nice additions to the documentation - thank you!

stencil-library/src/components/business-list/business-list.tsx
TiagoFuelber: Agree with that!
@jakemer 
What are your thoughts about using just a function that will read the snake case term and convert it to normal capitalized text? 
in`utils.ts` we have `snakeCaseToHumanReadable` function that does the work.

PR #220, opened 2023-11-02
Business form save on each step

jakemer:
- If a business id is passed, fetch the business and prefill the form
- If a business id was passed and the fetch was successful, PATCH on all steps
- If a business id was NOT passed, POST on the first step, store the business id that comes back from the request, and PATCH on all subsequent steps
- When POST/PATCH requests are taking place, disable the "next"/"previous"/"submit" buttons


TiagoFuelber (approved):
Great work!

PR #226, opened 2023-11-10
Make pagination consistent with customer admin

### LINKED ISSUE #198: Make component pagination consistent with customer-admin-fe
The web component library had an API file consistent with the customer-admin-fe app until the introduction of pagination into the component library. We should update the component library pagination component and API file to be consistent with how the dashboard does it.

- [pagination component example](https://github.com/justifi-tech/customer-admin-fe/blob/main/src/shared/components/tables/PaginationComponent.tsx#L29)
- [applying params example](https://github.com/justifi-tech/customer-admin-fe/blob/main/src/shared/api/Api.ts#L52)

In short, I don't believe the API file should have to worry about managing and formatting the pagination params - it should just apply all params that are sent along to it.

colganmeanor:
### Make pagination consistent with customer admin

This PR creates a new pagination-menu component that can be re-used in future list components, and it handles most of the pagination logic itself, which means the Api.ts file no longer has to handle this logic. 

Links
-----

closes #198 

Developer considerations
--------------

Steps for Testing/QA
--------------------

TBD


TiagoFuelber (approved):
Nice refactor!! Approved

PR #233, opened 2023-11-22
fix: handle tokenize error

### LINKED ISSUE #230: Fix: Error Details Not Passed Down on Payment Card Tokenization Failure
**Issue**:
We have received feedback from a client implementing the PaymentForm component. The client reported that when the card tokenization fails, the error_detail is not being passed down. This issue seems to be stemming from the way the response from the tokenization process is handled in the submit method of the PaymentForm component. Currently, the component emits the submitted event with the response of the tokenize method, but does not explicitly handle error scenarios or pass down error details.

**Impact**:
This issue prevents end-users from receiving meaningful feedback when a card tokenization error occurs, affecting user experience and potentially leading to confusion and decreased trust in the payment process.

**Proposed Solution**:
Modify the submit method in the PaymentForm component to include error handling. Specifically, after receiving the response from the tokenize method, the component should check if the response indicates a successful tokenization or an error. In case of an error, the component should emit an event with the error details. This will allow consuming components or applications to appropriately display error messages to end-users.

**Steps to Reproduce**:
Trigger the submit method in a scenario where card tokenization fails.
Observe that the error_detail is not passed down to the consuming component or application.

TiagoFuelber:
resolves #230 

* Add try catch block to handle the error response and emit error details along with the event

Links
-----

Developer considerations
--------------

Steps for Testing/QA
--------------------
* When submitting the `justifi-payment-form` and getting an error as response from tokenization, the `submitted` event should provide error information.

### DIFF
--- a/stencil-library/CHANGELOG.md
+++ b/stencil-library/CHANGELOG.md
@@ -1,5 +1,7 @@
 ### Changelog
 
+- Added Error handling for tokenization on PaymentForm
+
 ### 4.6.0
 
 > 23 October 2023
@@ -27,7 +29,6 @@
 - Updated table pagination to use proper bootstrap pagination styles
 - Updated `PaymentsList` and `PayoutsList` table headers and status badges to have descriptions on hover
 - Updated `PaymentsList` and `PayoutsList` to emit events when a table row is clicked
-
 
 #### 4.4.0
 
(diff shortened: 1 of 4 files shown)


nhalm:
For reference, the former response:
```json
{
    "status": 500,
    "error": "Internal Server Error",
    "exception": "<<long HTML here>",
    "traces": {
        "Application Trace": [
            {
                "exception_object_id": 33620,
                "id": 12,
                "trace": "app/services/capital_api.rb:42:in `get'"
            },
... LONG ARRAY CONTINUES
        ]
    }
}
```

New Response:
```json
{
    "id": "",
    "type": "payment_method",
    "error": "Internal Server Error",
    "page_info": "Error encountered during tokenization"
}
```

PR #234, opened 2023-11-27
201 new component refund form

TiagoFuelber:
image

* Create the Refund Form component.
* Still needs to test the API call and add some unit-tests
(Note: Not making real requests yet, waiting for definition about the proxy api)

Links
-----

Developer considerations
--------------

  - [X] This is a new MINOR version (functionality was added in a backwards compatible manner)

Steps for Testing/QA
--------------------

* Go to http://localhost:6006/?path=/story/pre-release-components-refundform--basic (local or staging)
* Set the `authToken` and `paymentId` props
* It should submit a refund request.
* Toggling the `withButton` prop should show a separate button that is set outside the component (the button will be displayed at the left side to differentiate from the default button)
* Filling in the  `refundInfoText` prop will display an information section above the form

### DIFF
--- a/stencil-library/src/components/refund-form/refund-form-schema.ts
+++ b/stencil-library/src/components/refund-form/refund-form-schema.ts
@@ -1,0 +1,18 @@
+import * as yup from 'yup';
+
+export interface RefundFormFields {
+  amount: number | null; // Allowing null for initial state or cleared inputs
+  message: string;
+}
+
+const RefundFormSchema = yup.object().shape({
+  amount: yup
+    .number()
+    .required('Amount is required')
+    .typeError('Amount must be a number')
+    .positive('Amount must be positive')
+    .nullable(), // To allow null values, useful if you clear the input
+  message: yup.string().trim().max(500, 'Notes must be 500 characters or less'), // Assuming a max length for notes
+});
+
+export default RefundFormSchema;
--- a/stencil-library/src/components/refund-form/refund-form.scss
+++ b/stencil-library/src/components/refund-form/refund-form.scss
@@ -1,0 +1,99 @@
+@import 'bootstrap-grid';
+@import 'bootstrap-utilities';
+
+:host {
+  @include host-base-styles;
+  display: block;
+}
+
+.info-text-container {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  gap: 1rem;
+  border-top: var(--jfi-form-control-border-color) solid 1px;
+  border-bottom: var(--jfi-form-control-border-color) solid 1px;
+  margin-bottom: 1rem;
+}
+
+.info-icon {
+  margin-left: 1rem;
+}
+
+.info-text {
+  flex: 1;
+  padding: 1rem;
+  font-size: 0.875rem;
+  line-height: 1.25rem;
+  color: var(--jfi-form-control-color);
+}
+
+*::part(input) {
+  font-family: var(--jfi-layout-font-family);
+  background-color: var(--jfi-form-control-background-color);
+  border-color: var(--jfi-form-control-border-color);
+  border-width: var(--jfi-form-control-border-width);
+  border-bottom-width: var(--jfi-form-control-border-bottom-width);
+  border-left-width: var(--jfi-form-control-border-left-width);
+  border-right-width: var(--jfi-form-control-border-right-width);
+  border-top-width: var(--jfi-form-control-border-top-width);
+  border-radius: var(--jfi-form-control-border-radius);
+  border-style: var(--jfi-form-control-border-style);
+  box-shadow: var(--jfi-form-control-box-shadow);
+  color: var(--jfi-form-control-color);
+  font-size: var(--jfi-form-control-font-size);
+  font-weight: var(--jfi-form-control-font-weight);
+  line-height: var(--jfi-form-control-line-height);
+  margin: var(--jfi-form-control-margin);
+  padding: var(--jfi-form-control-padding);
+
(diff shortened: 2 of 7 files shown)


stencil-library/src/components/refund-form/refund-form-schema.ts
```diff
@@ -0,0 +1,18 @@
+import * as yup from 'yup';
+
+export interface RefundFormFields {
+  amount: number | null; // Allowing null for initial state or cleared inputs
+  message: string;
+}
+
+const RefundFormSchema = yup.object().shape({
```
jakemer, line 8: We can figure it out in the PR where we get the refund to actually submit, but we should validate that the refund amount is not higher than the payment amount on the payment being refunded (might need to pass that as a prop)

stencil-library/src/components/refund-form/refund-form.scss
```diff
@@ -0,0 +1,99 @@
+@import 'bootstrap-grid';
+@import 'bootstrap-utilities';
+
+:host {
+  @include host-base-styles;
+  display: block;
+}
+
+.info-text-container {
```
jakemer, line 9: I think all of the custom styles in here can be achieved using bootstrap grid, utilitly classes, and the [border classes](https://getbootstrap.com/docs/5.3/utilities/borders/#border)

stencil-library/src/components/refund-form/refund-form.scss
```diff
@@ -0,0 +1,99 @@
+@import 'bootstrap-grid';
```
jakemer, line 1: I think this import should be called `grid`? That's how it is in other components like [here](https://github.com/justifi-tech/web-component-library/blob/main/stencil-library/src/components/payment-form/payment-form.scss#L1C1-L1C18)

I think you will also need `reboot` and `buttons`

stencil-library/src/components/refund-form/refund-form.scss
```diff
@@ -0,0 +1,99 @@
...
+  padding: var(--jfi-form-control-padding);
+
+  &:focus {
+    color: var(--jfi-form-control-color-focus);
+    border-color: var(--jfi-form-control-border-color-focus);
+    box-shadow: var(--jfi-form-control-box-shadow-focus);
+  }
+}
+
+.btn.jfi-submit-button {
```
jakemer, line 57: Are we able just to use bootstrap `btn`? I know that in the `payment-form` we do something similar to this but I am not sure why it was done that way. We should use bootstrap directly and export anything we want customers to be able to override

stencil-library/src/components/refund-form/refund-form.tsx
```diff
@@ -0,0 +1,188 @@
...
+                this.handleInput(name, value)
+              }
+              error={this.refundFieldsErrors.notes}
+            ></form-control-text>
+          </div>
+          {this.withButton && (
+            <div class="form-group d-flex flex-row-reverse">
+              <button
+                type="submit"
+                class={`btn btn-primary ml-auto jfi-submit-button${
```
jakemer, line 176: When submitting let's just make the button `disabled`

stencil-library/src/components/refund-form/refund-form.scss
TiagoFuelber: Perfect, fixed!

stencil-library/src/components/refund-form/refund-form.scss
TiagoFuelber: You mean, the button would still be styled with the jfi css variables, but the selector would be just `.btn`?

jakemer: I mean that we should use the `.btn` bootstrap class to apply styles for the button, and if a customer wants to customize it we can use the css exports. I think the payment form should have done that as well, and we would need less variables. Maybe there is a reason it was done this way though

stencil-library/src/components/refund-form/refund-form.tsx
```diff
@@ -0,0 +1,200 @@
...
+  /**
+   * Validates the form fields against the RefundFormSchema.
+   * Updates the refundFieldsErrors state with any validation errors.
+   */
+  private async validate() {
+    const newErrors = {};
+    let isValid = true;
+
+    try {
+      await RefundFormSchema.validate(this.refundFields, {
```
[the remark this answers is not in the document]
TiagoFuelber: Agree!

stencil-library/src/components/refund-form/refund-form-schema.ts
```diff
@@ -12,10 +12,10 @@ const RefundFormSchema = yup.object().shape({
     .typeError('Amount must be a number')
     .positive('Amount must be positive')
     .test(
-      'maxAmount',
+      'amount',
       'Refund amount cannot be more than the original payment amount',
       function (value) {
-        return value <= this.options.context.originalPaymentAmount;
+        return value <= this.options.context.amount;
```
jakemer, line 18: Makes sense why context needs to be passed to the validate function 👍

stencil-library/src/components/business-form/business-generic-info/business-generic-info.tsx
```diff
@@ -24,7 +24,6 @@ import { PHONE_MASKS } from '../../../utils/phone-masks';
 export class BusinessGenericInfo {
   @Prop() formController: FormController;
   @State() errors: any = {};
-  @State() business: any = {};
```
jakemer, line 27: Nice, thanks for doing this cleanup

PR #269, opened 2023-12-21
Env variables 

jakemer:
- Create a [`config.ts`](https://github.com/justifi-tech/web-component-library/pull/269/files#diff-56ea710ee5de5ca4c2a2fa37861394f8f0358aac455a31006be98e8c86742c39R1) that has keys of all env variables in it, and string values that are to be replaced by rollup
- Create an [`env.ts` ](https://github.com/justifi-tech/web-component-library/pull/269/files#diff-75726cad35531d8d3c2cc713b274787d80311d125ff082492b83c1daf9b0b788R7) that defaults to production env values, but if it can successfully parse an `.env` file it will use the values from that instead. The resulting env variables are then converted to a mapping used by the rollup replace plugin to replace the strings in `config.ts`
- Use `config.ts` throughout the code for getting variables - this way TS will actually catch when there is a typo or a variable is no longer defined (doesn't happen with using `process.env` directly because it must be type `any)

Developer considerations
--------------

Steps for Testing/QA
--------------------

### DIFF
--- a/stencil-library/config.ts
+++ b/stencil-library/config.ts
@@ -1,0 +1,13 @@
+export const config = {
+  iframeOrigin: 'ENV_IFRAME_ORIGIN',
+  exampleBusinessId: 'ENV_EXAMPLE_BUSINESS_ID',
+  exampleBusinessAccountId: 'ENV_EXAMPLE_BUSINESS_ACCOUNT_ID',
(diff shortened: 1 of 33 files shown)


TiagoFuelber (approved):
Nice work Jake! Approved

PR #280, opened 2024-01-05
fix: text wrapping on DetailItem

TiagoFuelber:
**New screenshot after changes (display table)**
image

* This adds some bootstrap classes to break long texts.
* Some sections (owners and additional questions) were not being displayed because of undefined values breaking the render. This adds optional chaining to all displayed values.
* This reorganizes the Owners section to display the owner's address beside the owner's details. (Let me know what you think). Also changed the Owner section to use the Owner's Name for clarity.

**Before**
image

 
**After**
image

**(NOTE)** Didn't fix the problem of a long string without spaces like this:
image

Ideas are welcome. One would be adding `overflow: auto` to the field, so the user could side scroll to see the whole string.

Links
-----

Developer considerations
--------------

Steps for Testing/QA
--------------------

### DIFF
--- a/stencil-library/src/components/details/utils.tsx
+++ b/stencil-library/src/components/details/utils.tsx
@@ -33,7 +33,7 @@ export const DetailSection = (
   <div part="detail-section" class="mt-4">
     <h5 part="detail-section-title">{sectionTitle}</h5>
     <hr />
-    <div class="d-flex flex-column justify-content-center gap-2 text-nowrap">
+    <div class="d-flex flex-column justify-content-center gap-2">
       {children}
     </div>
   </div>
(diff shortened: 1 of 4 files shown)


stencil-library/src/components/details/utils.tsx
```diff
@@ -33,7 +33,7 @@ export const DetailSection = (
   <div part="detail-section" class="mt-4">
     <h5 part="detail-section-title">{sectionTitle}</h5>
     <hr />
-    <div class="d-flex flex-column justify-content-center gap-2 text-nowrap">
```
jakemer, line 36: Can you try making using some of these utility classes to make the key/value pair sizing behave like an HTML table?
https://getbootstrap.com/docs/5.3/utilities/display/

So this line would be `.d-table`, the children would be `.d-table-row`, and their children (ln51 and ln54) would be `.d-table-cell`.

I think that will fix some of the weird alignment issues with the key/value pairs (you may need to import those utility classes)

jakemer (approved):

PR #281, opened 2024-01-05
Adjust table component responsiveness

### LINKED ISSUE #275: Responsive tables (*List components)
Currently our tables are not very responsive - at medium / small viewports text wraps or extends into neighboring columns. Check how `customer-admin-fe` handles table layout and scrolling.
- When tables don't fit the viewport they should scroll horizontally
- Whitespace no-wrap on certain columns

colganmeanor:
### Adjust table component responsiveness

Links
-----

closes #275 

Developer considerations
--------------

Steps for Testing/QA
--------------------

### DIFF
--- a/stencil-library/src/components/table/table.scss
+++ b/stencil-library/src/components/table/table.scss
@@ -14,7 +14,3 @@
 .table-wrapper {
   overflow-x: scroll;
 }
-
-.table-light {
-  white-space: nowrap;
-}
--- a/stencil-library/src/components/table/table.scss
+++ b/stencil-library/src/components/table/table.scss
@@ -8,4 +8,13 @@
 
 :host {
   @include host-base-styles;
+  table-layout: fixed;
+}
+
+.table-wrapper {
+  overflow-x: scroll;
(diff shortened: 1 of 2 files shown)


stencil-library/src/components/table/table.scss
```diff
@@ -8,4 +8,13 @@
...
 :host {
   @include host-base-styles;
+  table-layout: fixed;
 }
+
+.table-wrapper {
+  overflow-x: scroll;
+}
+
+.table-light {
```
TiagoFuelber, line 18: I think we could try to use a bootstrap class like `.text-nowrap` here. Or create a new class to add this style, the reason being that bootstrap uses `table-light` to style the table, and if we decide to change to `table-dark` we will have to update the CSS too.

colganmeanor: Great call - thanks for the suggestion, and explanation! 

Just updated it to simply apply the `.text-nowrap` class as you suggested, no need to create a new class here I think, or modify `table-light`.

PR #302, opened 2024-01-17
fix: added TS ignore comment on withActions type error

TiagoFuelber:
Links
-----

Developer considerations
--------------

Steps for Testing/QA
--------------------

### DIFF
--- a/apps/docs/stories/components/payments-list.stories.tsx
+++ b/apps/docs/stories/components/payments-list.stories.tsx
@@ -24,6 +24,7 @@ const meta: Meta = {
   },
   decorators: [
     customStoryDecorator,
+    // @ts-ignore - Ignore Storybook bug (reference to bug issue)
     withActions // https://github.com/storybookjs/storybook/issues/22384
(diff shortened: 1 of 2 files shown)


apps/docs/stories/components/payments-list.stories.tsx
```diff
@@ -24,6 +24,7 @@ const meta: Meta = {
   },
   decorators: [
     customStoryDecorator,
+    // @ts-ignore - Ignore Storybook bug (reference to bug issue)
```
jakemer, line 27: Nice! Is it possible to do at a more global level?

TiagoFuelber: I think so, but I am not sure if that would be good. It would bypass all "type mismatch" errors, which is quite common.

jakemer: Ahh, I see. Good call 👍

jakemer (approved):

PR #331, opened 2024-02-02
refactor(PaymentsList): decouple api calls to improve testability

TiagoFuelber:
## Overview
This PR introduces a significant refactor of the `PaymentsList` and `PayoutsList` components to enhance testability and maintainability. It separates the presentation layer from the logic, facilitating easier unit testing and future enhancements.

## Changes
- Refactored `PaymentsList` and `PayoutsList` to accept only props and inject dependencies into core components.
- Implemented dependency injection pattern to decouple business logic from UI components, improving test coverage.
- Updated unit tests to reflect the new component structure and added new tests for dependency logic.

* For a clearer understanding of the changes introduced in this file, please refer to the following articles:

[Scalable Frontend #1: Architecture](https://blog.codeminer42.com/scalable-frontend-1-architecture-9b80a16b8ec7/)
[Scalable Frontend #2: Common Patterns](https://blog.codeminer42.com/scalable-frontend-2-common-patterns-d2f28aef0714/)
[Scalable Frontend #3: The State Layer](https://blog.codeminer42.com/scalable-frontend-3-the-state-layer-b23ed69ca57c/)

Links
-----

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA 
- On new features 
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation

Developer QA steps
--------------------


packages/webcomponents/src/components/table/table.tsx
```diff
@@ -80,6 +86,7 @@ export class Table {
                         this.rowData?.map((data, index) => {
                           return (
                             <tr
+                              data-test-id="payment-row"
```
jakemer: Should we pass test ids through as props? Are there others?

packages/webcomponents/src/components/payments-list/fetch-payments-response.type.ts
```diff
@@ -0,0 +1,96 @@
...
+  acct_last_four: string;
+  brand: string;
+  name: string;
+  token: string;
+  metadata: null;
+  created_at: string;
+  updated_at: string;
+}
+
+interface BankAccount {
```
jakemer, line 19: I believe we have a lot of these types already and they will be needed for other components, does it make sense to put the ones that are missing in the files in the `api` directory?

packages/webcomponents/src/components/table/table.tsx
TiagoFuelber: Good one, I renamed it to `table-row`

packages/webcomponents/src/api/Payment.ts
```diff
@@ -284,3 +299,30 @@ export class PaymentBalanceTransaction implements IPaymentBalanceTransaction {
...
+
+export interface IApplicationFee {
+  id: string;
+  amount: number;
+  currency: string;
+  created_at: string;
+  updated_at: string;
+}
+
+export interface FetchPaymentsResponseType {
```
colganmeanor, line 323: I am curious - what is the reason we created a new type interface for the Payment List response, instead of using `IApiResponse` or `IApiResponseCollection` from `Api.ts`? 

It contains the exact properties you have specified here. 🤔

packages/webcomponents/src/components/payments-list/get-payments.action.ts
```diff
@@ -0,0 +1,38 @@
+import { Payment } from '../../api';
+
+export const getPayments = async ({
```
colganmeanor, line 3: @TiagoFuelber I've been looking at this example when trying to recreate this type of logic for some other components, I think there is one problem here that causes a wrinkle or two. 

Mainly, the props that are getting passed to `getPayments` are not really typed as we expect. In this current implementation, we call `getPayments()` and pass it an object that contains all the values from the `payments-list-core` file. But nothing in this object is assigned a type definition, so this function is assuming all of the types for the values inside that object are `any`. 

I notice this especially when looking at the type of `response` on line 17 of this file. `response` is typed as `any`, and so TS doesn't throw an error on line 18. If `service` is typed as `PaymentService` which I believe would be the intention, then we would start to get TS errors whenever we try to reference `response.error` because it doesn't exist on the `FetchPaymentsResponseType` that was created. 

So, in conjunction with the other comment I left, I think I'd suggest removing the `FetchPaymentsResponseType` and using `IApiResponse` in it's place. What do you think? 🤔

packages/webcomponents/src/components/payments-list/payment.service.ts
```diff
@@ -0,0 +1,22 @@
...
+    params: any
+  ): Promise<FetchPaymentsResponseType>;
+}
+
+export class PaymentService implements IPaymentService {
+  async fetchPayments(
+    accountId: string,
+    authToken: string,
+    params: any
+  ): Promise<FetchPaymentsResponseType> {
```
colganmeanor, line 17: If you decide to implement the changes in my other comments - just need to make sure to update: `Promise:<FetchPaymentsResponseType>` in this file to `Promise<IApiResponseCollection<IPayment[]>>`

packages/webcomponents/src/api/Payment.ts
TiagoFuelber: Hey, nice catch! @colganmeanor 

I mislooked that we had these Api response types already. It makes total sense to use them. I am changing this to use our existing response types. 👍

packages/webcomponents/src/api/Payment.ts
```diff
@@ -148,24 +149,32 @@ export interface IPayment {
...
   amount_refunded: number;
+  amount_returned?: number;
   balance: number;
   captured: boolean;
   capture_strategy: CaptureStrategy;
   currency: 'usd';
   description: string;
   disputed: boolean;
-  disputes: IDispute[];
+  disputes: (IDispute | null)[];
```
colganmeanor, line 159: Is it possible to type `disputes` this way? It reads to me like `disputes` will always be a truthy array with either `IDispute` or `null` as items in the array. I know its not possible to have `null` index items in the array

Could it be: `IDispute[] | null` ? 🤔

packages/webcomponents/src/components/table/utils.tsx
```diff
@@ -18,8 +18,8 @@ export const EmptyState = (columnSpan: number) => (
 
 export const ErrorState = (columnSpan: number, errorMessage: string) => (
   <tr>
-    <td class="error-state" part="error-state" colSpan={columnSpan} style={{ textAlign: 'center' }}>
+    <td class="error-state" part="error-state" colSpan={columnSpan} data-test-id="error-state" style={{ textAlign: 'center' }}>
```
colganmeanor, line 21: Very small thing - since this error state is specifically for the table, could we rename this to `data-test-id='table-error-state'` ?

packages/webcomponents/src/components/pagination-menu/test/pagination-menu.spec.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    // Convert classList to a string for the assertion
+    const previousButtonClasses = previousButton.className;
+    const nextButtonClasses = nextButton.className;
+
+    expect(previousButtonClasses).toContain('disabled');
+    expect(nextButtonClasses).not.toContain('disabled');
+  });
+
+  it('emits events on button clicks', async () => {
```
colganmeanor, line 81: this is really nice to see event calls being tested in our spec tests - good work here!

packages/webcomponents/src/components/payments-list/get-payments.ts
```diff
@@ -0,0 +1,29 @@
+import { Payment } from '../../api';
+
+export const makeGetPayments =
```
colganmeanor, line 3: can this function be renamed to `getPayments` ? 

I'm finding `makeGetPayments` a little confusing to read

packages/webcomponents/src/components/payments-list/payments-list-core.tsx
```diff
@@ -0,0 +1,122 @@
...
+  @State() paging: PagingInfo = pagingDefaults;
+  @State() params: any;
+  @Event({
+    eventName: 'payment-row-clicked',
+
+    bubbles: true,
+  }) rowClicked: EventEmitter<Payment>;
+
+  componentWillLoad() {
+    if (typeof this.getPayments === 'function') {
```
colganmeanor, line 24: I'm not totally sure I understand this line 24 - is this so that test spec files won't try to call `fetchData`? 🤔

packages/webcomponents/src/components/payments-list/payments-list-core.tsx
```diff
@@ -0,0 +1,122 @@
...
+  handleClickNext = (afterCursor: string) => {
+    const newParams: any = { ...this.params };
+    delete newParams.before_cursor;
+    this.params = ({ ...newParams, after_cursor: afterCursor });
+  };
+
+  fetchData(): void {
+    this.loading = true;
+
+    if (typeof this.getPayments === 'function') {
```
colganmeanor, line 50: Similar to my other comment in this file - I think this logic is repeated here when perhaps it doesn't need to be?

colganmeanor (commented):
I left a few comments and questions. It's a bigger PR so I think I'm going to take another look at it tomorrow morning to make sure I understand it all before approving. 

Overall, I think your test files look great, and the added layers of separation and abstraction you've added have contributed to making the test files that much better. Great job! 

I'm wondering if we can find any ways to improve dev readability and make it a bit more clear and concise. 🤔

packages/webcomponents/src/components/payments-list/payments-list-core.tsx
TiagoFuelber: This is just to prevent `fetchData` from being called in case no `getPayments` is passed (which is only in test files) 

This can be simplified to just `if (this.getPayments)`. I'll change that and also remove the redundant validation on line 50. Good catch!

packages/webcomponents/src/components/payments-list/get-payments.ts
TiagoFuelber: Thanks for the feedback! makeGetPayments is actually a factory function, hence the 'make' prefix—it sets up a custom function with specific settings. I understand the confusion, though. What do you think about `createPaymentFetcher` as an alternative? It might strike a better balance.

colganmeanor: Ah I see - thank you for explaining @TiagoFuelber ! 

I think I like `createPaymentFetcher` a little better, but in that case I will let you decide since its a factory function. I can live with either lol.

packages/webcomponents/src/api/Payment.ts
TiagoFuelber: Oh good catch, this is not needed. I changed that during development to test something and forgot to undo.
Thanks!

colganmeanor (approved):
🍕

PR #335, opened 2024-02-07
refactor(PaymentDetails): create payment-details-core component

### LINKED ISSUE #344: Refactor component render logic and improve testability: PaymentDetails
- Create a service.ts file to handle the API request type logic.
- Create mockService.ts file that implements interface from service.ts
- Create a mockSuccessResponse.json file
- Update spec test logic to mockService.ts API with mockResponse.json

TiagoFuelber:
* This creates the `payment-details-core` component receiving a payment service, and adds some unit tests with snapshots

Links
-----

closes #344 

Developer considerations
--------------

- On any task
  - [x] Perform dev QA
- On new features
  - [x] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation

Developer QA steps
--------------------


packages/webcomponents/src/components/payment-details/get-payment-details.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const makeGetPaymentDetails =
+  ({ id, authToken, service }) =>
+  async ({ onSuccess, onError }) => {
+    try {
+      const response = await service.fetchPayment(id, authToken);
+      if (!response.error) {
+        onSuccess({ payment: new Payment(response.data) });
+      } else {
+        const responseError =
```
jakemer, line 11: Can we make a helper for this so that we don't have the possibility of handling response errors differently across components/services? Can we abstract even more of this so that the services are quick and easy to write, and consistent? It could be a separate follow-up PR after your pending PRs are closed

packages/webcomponents/src/components/payment-details/test/__snapshots__/payment-details-core.spec.tsx.snap
```diff
@@ -0,0 +1,186 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
jakemer, line 1: Ahh, I see how these snapshots work now. Very cool.

packages/webcomponents/src/components/payment-details/payment-details-core.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  componentWillLoad() {
+    if (typeof this.getPaymentDetails === 'function') {
+      this.fetchData();
+    }
+  }
+
+  @Watch('paymentId')
+  @Watch('authToken')
+  @Watch('paymentService')
```
jakemer, line 25: I think since `paymentId`, `authToken`, and `paymentService` are not props or pieces of state this will never fire. However, I don't think it's necessary to watch for any changes here anyways. `payment-details.tsx` watches for `paymentId` and `authToken` changes then re-initializes and updates `getPaymentDetails` which is in state. That should cause a re-render, which should cause this component to be re-initialized

packages/webcomponents/src/components/payment-details/payment-details.tsx
```diff
@@ -30,111 +29,49 @@ import { config } from '../../../config';
...
   @Prop() authToken: string;
-  @State() payment: Payment;
-  @State() loading: boolean = true;
-  @State() errorMessage: string;
 
-  @Watch('paymentId')
-  @Watch('authToken')
-  updateOnPropChange() {
-    this.fetchData();
+  @State() getPaymentDetails: Function;
```
jakemer, line 39: I'm not sure how `@State` works with a function, but I do know that deep comparisons don't usually take place when Stencil is deciding to do a re-render. I think since `initializeGetPaymentDetails` creates a _new_ function using `makeGetPaymentDetails` it should cause a re-render. You may want to double check if changing `paymentId` or `authToken` causes a re-render and for `payment-details-core` to be re-initialized just to be sure

packages/webcomponents/src/components/payment-details/test/payment-details.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+describe('payment-details', () => {
+  const components = [PaymentDetails, PaymentDetailsCore, Details];
+  it('renders an error message when paymentId and authToken are not provided', async () => {
+    const page = await newSpecPage({
+      components,
+      html: '<justifi-payment-details></justifi-payment-details>',
+    });
+    await page.waitForChanges();
+    expect(page.root).toMatchSnapshot();
```
jakemer, line 14: Very interesting - it took me a minute to get how this was working because the `toMatchSnapshot()` seemed a little unclear at first, but I see now that the snapshot holds a bunch of states keyed by the `it` statement. Nice.

jakemer (dismissed):
Wow 🤩 ! Fantastic work on this @TiagoFuelber . I love the separation of concerns. Not only does it make things wayyy more testable, it also just feels more compartmentalized and clean

packages/webcomponents/src/components/payment-details/test/payment-details.spec.tsx
TiagoFuelber: Exactly, it's very helpful. We can catch errors easily with this.

packages/webcomponents/src/components/payment-details/payment-details.tsx
TiagoFuelber: Thanks for pointing that out! You're right about `@State()` and function references. In our case, since `initializeGetPaymentDetails` assigns a new function, it does trigger a re-render, ensuring `payment-details-core` updates as expected with `paymentId` or `authToken` changes. This aligns with our real-world usage where the component behaves as intended. I'll double-check to confirm the behavior remains consistent. Appreciate the insight!

packages/webcomponents/src/components/payment-details/get-payment-details.ts
TiagoFuelber: Absolutely, that's a great idea! I've included the helper in the PR to ensure consistency across our components and services. Thanks for the suggestion!

colganmeanor (approved):
🧋

PR #336, opened 2024-02-07
refactor(PayoutDetails): create payout-details-core component

### LINKED ISSUE #346: Refactor component render logic and improve testability: PayoutDetails
- Create a service.ts file to handle the API request type logic.
- Create mockService.ts file that implements interface from service.ts
- Create a mockSuccessResponse.json file
- Update spec test logic to mockService.ts API with mockResponse.json

TiagoFuelber:
* This creates the `payout-details-core` component and adds unit tests

**Snapshot testing**
```
it('renders loading state initially', async () => {
    const page = await newSpecPage({
      components: [PayoutDetailsCore],
      html: '<payout-details-core></payout-details-core>',
    });
    expect(page.root).toMatchSnapshot();
  });
  ```
 This test checks the initial loading appearance of a component named PayoutDetailsCore. It does this by creating a test page with the component and then taking a "snapshot" of the component's HTML structure. This snapshot is saved in a file with a .snap extension.

If later changes to the component unintentionally alter its HTML structure, this test will detect the differences by comparing the new HTML output against the saved snapshot. If there's a mismatch, the test will fail, signaling that something has changed.

If the changes to the HTML are deliberate, you can update the snapshot to the new version by pressing u while the test suite is running in your terminal. This will accept the changes and save the new HTML structure as the expected snapshot.

Links
-----

closes #346 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation

Developer QA steps
--------------------


colganmeanor (dismissed):
🧁

colganmeanor (approved):

PR #342, opened 2024-02-08
Update BusinessForm Logic - Patch Only Requests

### LINKED ISSUE #327: BusinessForm - require business-id and only PATCH
Our intention for the BusinessForm is to update businesses during onboarding but to create them via API before that. As such, we should:
- Remove logic that does POST requests to create a business
- Require `business-id` to be passed to the component
- Consider what the state should be when a `business-id` is not passed or cannot be loaded
- Disabled "owners" form section if there is data. If not, new owners can be added and patched. In the `PATCH` request, if the `owners` object is there, don't send it again

colganmeanor:
### Update BusinessForm Logic - Patch Only Requests

- `businessId` is now a required prop on `business-form.tsx` and `business-form-stepped.tsx`
- added a `FormAlerts` for IF `businessId` or `auth-token are missing`
   - disabled `Submit` and `Next` buttons IF either of these props are missing, or if there is a server error. 
- `SendData` requests in both components - removed `POST` functionality and trimmed down functions to only send `PATCH` requests
- added logic to check for `existingOwners` when fetching business data
  - IF `existingOwners` is truthy: 
    - Form inputs are DISABLED in `owners-form`
    - Owner info is NOT sent in data payload
  - ELSE
    - Form inputs are enabled in `owners-form`
    - Owner info is sent in data payload for patching in new `Owners`

Had to disable a test in the business-form.spec file that no longer made sense, harder to test at the moment. 

Links
-----

closes #327 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation

Developer QA steps
--------------------


packages/webcomponents/src/components/business-form/business-form-stepped.tsx
```diff
@@ -55,16 +43,31 @@ export class BusinessFormStepped {
...
+    0: (formController) => <justifi-business-generic-info formController={formController} />,
+    1: (formController) => <justifi-legal-address-form formController={formController} />,
+    2: (formController) => <justifi-additional-questions formController={formController} />,
+    3: (formController) => <justifi-business-representative formController={formController} />,
+    4: (formController) => <justifi-business-owners formController={formController} />
+  };
+
   componentWillLoad() {
     if (!this.authToken) {
       console.warn(
```
jakemer, line 56: I think these should probably be errors instead of warnings since the component won't function without it

packages/webcomponents/src/components/business-form/business-form-stepped.tsx
```diff
@@ -26,25 +18,21 @@ const componentStepMapping = {
...
   @State() errorMessage: string = '';
 
-  get submitDisabled() {
-    return !this.authToken || this.isLoading || this.serverError;
-  }
-
-  get nextDisabled() {
-    return !this.authToken || this.isLoading || this.serverError;
+  get disabledState() {
+    return !this.authToken || !this.businessId || this.isLoading || this.serverError;
```
jakemer, line 31: I don't think this logic works because if a user attempts to save something that results in a server error and they correct it, they will not be able to try submitting again. I also think that we should leave `!this.authToken` and `!this.businessId` out and just let the console errors do the work to lower code complexity

packages/webcomponents/src/components/business-form/business-form-stepped.tsx
```diff
@@ -55,16 +43,31 @@ export class BusinessFormStepped {
...
       console.warn(
         'Warning: Missing auth-token. The form will not be functional without it.',
       );
+      this.errorMessage = 'Missing auth-token. The form will not be functional without it.';
+    }
+    if (!this.businessId) {
+      console.warn(
+        'Warning: Missing business-id. The form requires an existing business-id to function.'
+      )
+      this.errorMessage = 'Missing business-id. The form requires an existing business-id to function.';
```
jakemer, line 65: Let's not set user-facing errors here. An implementing dev should test their implementation and see the console errors but I don't think this should ever make it to the user

jakemer (changes requested):

PR #357, opened 2024-02-14
Business form submit

jakemer:
The `BusinessForm` was not submitting due to validation errors which were not being shown. You can see in the following image that two things were happening.
1) `legal_address.line2` was being required when it is optional data
2) `additional_questions`, which is a whole form section, was being validated. This error has no place in the UI nor do we want it to. This was taking precedence over it's individual form field validations which we _do_ want to show:
Screen Shot 2024-02-14 at 10 09 49 AM

To fix this we:
- Created a copy of `Business.ts`, named `BusinessV2.ts` with fallback values (`{}`) for sections
- Use `Business` class to create new business instance from loaded business response data
- Removed `required` validations from form sections so that they can be empty, which also allows individual form field input validations to be shown

Links
-----

Developer considerations
--------------

- On any task
  - [x] Perform dev QA
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation

Developer QA steps
--------------------

- [x] The form now shows validation errors for "Additional Questions"
- [x] The form will now submit when all fields are valid

### DIFF
--- a/packages/webcomponents/src/components/form/form.ts
+++ b/packages/webcomponents/src/components/form/form.ts
@@ -99,6 +99,8 @@ export class FormController {
     context?: any,
   ): Promise<void> {
     const isValid = await this.validate(context || {});
+    console.log('this._errors', this._errors);
+    console.log('isValid', isValid);
     if (isValid) {
       submitHandler();
     }
--- a/packages/webcomponents/src/components/form/form.ts
+++ b/packages/webcomponents/src/components/form/form.ts
@@ -99,8 +99,6 @@ export class FormController {
     context?: any,
   ): Promise<void> {
     const isValid = await this.validate(context || {});
-    console.log('this._errors', this._errors);
-    console.log('isValid', isValid);
     if (isValid) {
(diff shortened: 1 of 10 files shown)


packages/webcomponents/src/components/form/form.ts
```diff
@@ -99,6 +99,8 @@ export class FormController {
     context?: any,
   ): Promise<void> {
     const isValid = await this.validate(context || {});
+    console.log('this._errors', this._errors);
```
colganmeanor, line 102: lets remove these before re-submitting

colganmeanor (commented):
looks good overall - lets remove the business-form-stepped logic from here and these console logs in the handleValidate function

jakemer:
Additional question errors are now shown:
Screen Shot 2024-02-14 at 2 36 46 PM

And the form submits when the validation errors shown are addressed:
Screen Shot 2024-02-14 at 2 44 24 PM

colganmeanor (approved):

PR #359, opened 2024-02-14
BusinessForm "Country" input is disabled, required, and does not have a value

jakemer:
The `BusinessForm` requires a `business-id` for an existing business, which it will then load and pre-fill the form. However, when the form is pre-filled, the country is not being set.

This causes the form to not submit because country is required, disabled (because we only support USA as of now)  and has no value.

We should leave the validator and the disabled state and make sure the value of the select is set to the country saved in the loaded business, or defaults to "United States"

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation

Developer QA steps
--------------------
- Run the branch locally
- In Storybook go to "BusinessForm"
- [x] Under "Legal Address", the country selector is disabled and is displaying "United States"
- [x] Fill out / submit the form. The request data should contain `'USA'` for `legal_address.country`

### DIFF
--- a/.changeset/gentle-pots-whisper.md
+++ b/.changeset/gentle-pots-whisper.md
@@ -1,0 +1,5 @@
+---
+"@justifi/webcomponents": patch
+---
+
+Default `BusinessForm` legal address country to 'USA', and update select option value to 'USA' instead of 'US'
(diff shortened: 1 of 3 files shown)


colganmeanor (dismissed):
🦍

colganmeanor (approved):

PR #389, opened 2024-03-01
Fix country business form

### LINKED ISSUE #391: Fix business form country field for businesses created with legal_address with no country value
Our previous PR attached to #352 was not complete enough to fix this issue. 

While this fixed the issue for freshly created businesses with `legal_address: null`, it does not fix the issue for businesses created with other properties inside `legal_address`. 

For example -  a business created with only `legal_address: { "state": "${state_name}" }` will still encounter the issue with the form. 

The address class is not being properly instantiated in these cases, and the fallback data of `USA` is not being properly applied to the `country` field in all cases.

colganmeanor:
### Fix Country field in Business Class to correctly fill Business Form

Links
-----

Fixes #391 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------

- [ ] Test Business Form Loading and Submitting Data for businesses created in various states
  - [ ] Test full completed Business data in form
  - [x] test newly created business data with no `legal_address` data passed (`legal_address` should be null when starting the form
  - [x] test newly created business data with only SOME info passed to `legal_address`, but NOT `country`.

### DIFF
--- a/packages/webcomponents/src/components/business-form/business-generic-info/business-generic-info.tsx
+++ b/packages/webcomponents/src/components/business-form/business-generic-info/business-generic-info.tsx
@@ -99,7 +99,7 @@ export class BusinessGenericInfo {
                 name="industry"
                 label="Industry"
                 defaultValue={genericInfoDefaultValue.industry}
-                error={this.errors.business_structure}
+                error={this.errors.industry}
                 inputHandler={this.inputHandler}
               />
             </div>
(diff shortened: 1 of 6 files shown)


packages/webcomponents/src/components/business-form/business-generic-info/business-generic-info.tsx
```diff
@@ -99,7 +99,7 @@ export class BusinessGenericInfo {
                 name="industry"
                 label="Industry"
                 defaultValue={genericInfoDefaultValue.industry}
-                error={this.errors.business_structure}
+                error={this.errors.industry}
```
jakemer, line 102: Nice catch!

jakemer (approved):
Looks good to me

PR #393, opened 2024-03-05
Fix business details broken

### LINKED ISSUE #315: BusinessDetails component appears broken with freshly created businesses with only legal name
The rendering of the BusinessDetails web component is incorrect when the business has only a legal name. From the provided image, there might be a missing null check somewhere in the code.

Image

TiagoFuelber:
closes #315 

This PR enhances the BusinessDetails component by introducing validation logic to ensure only sections with available data are displayed, while sections lacking data are hidden. This improvement primarily addresses the scenario where a business is registered through the API with merely a legal name provided.

Key points:

* The validation logic currently targets the specific case of businesses created with only a legal name. Should additional edge cases emerge, further validations will be considered and implemented as needed.
* Legal Address Handling: A function named showFirstColumn has been introduced within the Legal Address section. This function prevents the Country field from being isolated in the second column. Given that the country field defaults to 'USA' for all businesses, it ensures the Legal Address section is not left empty.
* Future Consideration: We may explore the possibility of concealing the Legal Address section entirely when the only detail present is the country.

This update aims to improve the user experience by avoiding the display of incomplete or irrelevant sections within the BusinessDetails component.

Screenshot 2024-03-05 at 11 56 57

Links
-----

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation

Developer QA steps
--------------------
[ ] - Create a business with only legal name on it
[ ] - BusinessDetails should not display irrelevant sections


nhalm:
@TiagoFuelber 
I still have a not found when I run `pnpm run build && pnpm run dev` locally:

CleanShot 2024-03-05 at 09 22 26

packages/webcomponents/src/components/business-details/generic-info-details/generic-info-details.tsx
```diff
@@ -17,34 +17,47 @@ export class GenericInfoDetails {
...
-                value={snakeCaseToHumanReadable(this.business?.business_type)}
-              />
-              <DetailItem
-                title="Business Structure"
-                value={snakeCaseToHumanReadable(
-                  this.business?.business_structure,
-                )}
-              />
-              <DetailItem title="Industry" value={this.business?.industry} />
+              {this.business?.legal_name && (
```
colganmeanor, line 20: I'm not sure if these conditionals are needed here for the `DetailItem` components - inside `DetailItem` contains a ternary that renders the component OR null based on the presence of `title` and `value`

```
export const DetailItem = ({
  title,
  value,
}: {
  title: string;
  value: string;
}) =>
  title && value ? (
    <div class="d-table-row gap-2">
      <span part="detail-section-item-title" class="fw-bold d-table-cell px-2">
        {title}
      </span>
      <span
        part="detail-section-item-data"
        class="flex-1 d-table-cell px-2 text-wrap"
        innerHTML={value}
      ></span>
    </div>
  ) : null;
```

So maybe we can get away with doing this without these checks? What do you think?

packages/webcomponents/src/components/business-details/business-details.tsx
```diff
@@ -86,12 +87,20 @@ export class BusinessDetails {
     return (
       <Host>
         <generic-info-details business={this.business} />
-        <legal-address-details legalAddress={this.business.legal_address} />
-        <representative-details
-          representative={this.business.representative}
-        />
-        <owner-details owners={this.business.owners} />
-        <additional-questions-details additionalQuestions={this.business.additional_questions} />
+        {!isEmptyObject(this.business.legal_address) && (
```
colganmeanor, line 90: I think it's a good call to create this `isEmptyObject` function and export it from the utils - but I think it makes this file a little less clean. 

Would you consider this as an alternative? 

What if we moved this logic inside the `DetailSection` component or inside the sub components? Since each of these `business-details` sub components are rendering `DetailSection` - maybe we do the check for `isEmptyObject` inside the detail components, such as inside `representative-details`, and then pass that boolean to `DetailSection` where it uses the boolean in a ternary to render the section or not. 

What do you think of this possibilty?

TiagoFuelber: Very good call!

I'll move this validation to inside each subcomponent 🙌

colganmeanor (approved):
🐝

✅

PR #394, opened 2024-03-05
Stepped form subcomponents handle their own data

colganmeanor:
### Business Form Stepped sub components handle their own data

This PR refactors `business-form-stepped` so that each form sub component handles its own data (fetching data and patching data). It also removes the form controller from the parent component, and each form sub component now manages its own Form State.

Links
-----

Developer considerations
--------------

- On any task
  - [x] Perform dev QA

Developer QA steps
--------------------

- [x] Unit tests are passing
- [x] Verify that stepped form is now successfully loading and patching data on each form step. 
- [x] Please help try to break the form to expose bugs and logic gaps
   - [x] Load businesses with varying states of data completion into the form - IE an existing business with data fully filled out, a brand new business with just a `legal_name`, etc


packages/webcomponents/src/components/business-form/business-form-schema.ts
```diff
@@ -111,7 +111,7 @@ export const BusinessStructureOptions: {
     },
   ];
 
-export const businessGenericInfoSchema = object({
```
jakemer, line 114: Can we start breaking this up into files that can go alongside the appropriate components? And anything that is shared can stay here, though maybe it will make sense to change the file name at that point

packages/webcomponents/src/components/business-form-stepped/business-core-info/business-core-info-form-step.tsx
```diff
@@ -48,101 +117,103 @@ export class BusinessCoreInfoFormStep {
...
-            <div class="col-12 col-md-6">
-              <form-control-number-masked
-                name="phone"
-                label="Phone Number"
-                defaultValue={genericInfoDefaultValue.phone}
-                error={this.errors.phone}
-                inputHandler={this.inputHandler}
-                mask={PHONE_MASKS.US}
-              />
+        <form onSubmit={() => this.sendData}>
```
jakemer, line 124: I think this might not be necessary since the `validateAndSubmit` method will be called

packages/webcomponents/src/components/business-form-stepped/business-core-info/business-core-info-form-step.tsx
```diff
@@ -23,17 +28,81 @@ import { CoreBusinessInfo, ICoreBusinessInfo } from '../../../api/Business';
...
+    }
+  }
+  
+  handleResponse(response, onSuccess) {
+    if (response.error) {
+      this.serverError.emit({ data: response.error, message: 'Error updating business data' });
+    } else {
+      onSuccess();
+    }
+    this.submitted.emit({ data: response });
```
jakemer, line 84: Instead of having `this.serverError` and `this.submitted`, can we have JUST `this.submitted` and always `emit` the full response instead of formatting it differently? This way we don't have to worry about any inconsistencies, it's just always the response from the server

packages/webcomponents/src/components/business-form/business-form-schema.ts
colganmeanor: I was planning on doing another cleanup PR that would address things like this file, and making sure that we have items that are shared between `business-form` and `business-form-stepped` in their appropriate place. 

We can totally do it in this file if you prefer, but I wanted to keep this PR mostly focused on the form sub-components managing their own Form and data

packages/webcomponents/src/components/business-form-stepped/business-core-info/business-core-info-form-step.tsx
colganmeanor: Good call! Fixed

packages/webcomponents/src/components/business-form-stepped/business-core-info/business-core-info-form-step.tsx
colganmeanor: We can absolutely change it - 

However here is my logic for doing it this way: 

`this.serverError` is NOT emitted all the way to the customer. My intention with emitting it as an event this way is only for the parent component (`business-form-stepped`) to listen to it from the child components and trigger an Error state in the parent component. 

`this.submitted` is still the event that is emitted to the customer / dev from the sub-components, and it gets emitted regardless of a successful response OR an error response. 

Perhaps it is confusing and unnecessary to emit the full error object / data in the event, and we could simply emit the string that says `Failure to get Business Data` or `Failure to update business data`. 

What do you think?

packages/webcomponents/src/components/business-form-stepped/legal-address-form/legal-address-form-step.tsx
```diff
@@ -11,101 +16,164 @@ import { FormController } from '../../form/form';
...
+
+  get businessEndpoint() {
+    return `entities/business/${this.businessId}`
+  }
+
+  private async fetchData() {
+    this.formLoading.emit(true);
+    try {
+      const response: IApiResponse<IBusiness> = await this.api.get(this.businessEndpoint);
+      this.legal_address = response.data.legal_address;
```
TiagoFuelber, line 44: I think we could use the Address class here, this way we'll always have a country set. 
I tried to edit a business created from Insomnia, and got stuck on the second step because the business didn't have a country.

colganmeanor: Good call - thanks man!

packages/webcomponents/src/components/business-form-stepped/business-core-info/business-core-info-form-step.tsx
colganmeanor: Discussed this again in standup this morning, we are going to use the event emitter logic here to trigger the form alert in the parent form component.

packages/webcomponents/src/components/business-form-stepped/business-form-stepped.tsx
```diff
@@ -21,25 +16,16 @@ export class BusinessFormStepped {
...
-  private api: any;
 
   constructor() {
-    this.sendData = this.sendData.bind(this);
-    this.fetchData = this.fetchData.bind(this);
-  }
-
-  get disabledState() {
-    return this.isLoading;
+    this.incrementSteps = this.incrementSteps.bind(this);
```
TiagoFuelber, line 27: we can skip binding `this` if we use arrow functions when declaring the function, example: 

line 103.
`incrementSteps = () => {}`

colganmeanor: Oh I did not know this - thanks for the tip @TiagoFuelber ! 

Updated this!

packages/webcomponents/src/components/business-form-stepped/business-form-stepped.tsx
```diff
@@ -21,26 +16,12 @@ export class BusinessFormStepped {
   @Prop() businessId: string;
   @Prop() testMode: boolean = false;
   @Prop() hideErrors?: boolean = false;
-  @State() isLoading: boolean = false;
+  @State() formLoading: boolean = false;
+  @State() serverError: boolean = false;
```
TiagoFuelber, line 20: We could consider checking if `errorMessage` has a message, then this state wouldn't be necessary.

packages/webcomponents/src/components/business-form-stepped/business-core-info/business-core-info-form-step.tsx
```diff
@@ -23,17 +24,81 @@ import { CoreBusinessInfo, ICoreBusinessInfo } from '../../../api/Business';
...
   @State() errors: any = {};
-  @State() genericInfo: ICoreBusinessInfo = {};
+  @State() coreInfo: ICoreBusinessInfo = {};
+  @Event({ bubbles: true }) submitted: EventEmitter<{ data?: any}>;
+  @Event() formLoading: EventEmitter<boolean>;
+  @Event() serverError: EventEmitter<{ data: any, message: BusinessFormServerErrors }>;
 
   constructor() {
     this.inputHandler = this.inputHandler.bind(this);
+    this.sendData = this.sendData.bind(this);
```
TiagoFuelber, line 38: Here we could use arrow functions as well

packages/webcomponents/src/components/business-form-stepped/business-form-stepped.tsx
```diff
@@ -63,137 +74,98 @@ export class BusinessFormStepped {
...
 
   previousStepButtonOnClick() {
     this.clickEvent.emit({ name: ClickEvents.previousStep })
-    this.sendData(() => this.currentStep--);
-  }
-
-  showNextStepButton() {
-    return this.currentStep < this.totalSteps;
+    const currentStep = this.refs[this.currentStep];
+    currentStep.validateAndSubmit(this.decrementSteps);
```
TiagoFuelber, line 108: This could be slightly easier to read if we do: 
`currentStep.validateAndSubmit({ onSuccess: this.decrementSteps });`

And then in `business-core-info-form-step.tsx`
```
  @Method()
  async validateAndSubmit({ onSuccess }: { onSuccess: () => Promise<void> }) {
    this.formController.validateAndSubmit(() => this.sendData(onSuccess));
  };
  ```

File: packages/webcomponents/src/components/business-form-stepped/business-core-info/business-core-info-form-step.tsx [hunk 9]
```diff
@@ -23,17 +24,81 @@ import { CoreBusinessInfo, ICoreBusinessInfo } from '../../../api/Business';
...
+    if (response.error) {
+      this.serverError.emit({ data: response.error, message: BusinessFormServerErrors.patchData });
+    } else {
+      onSuccess();
+    }
+    this.submitted.emit({ data: response });
+  }
+
+  @Method()
+  async validateAndSubmit(onSuccess: () => Promise<void>) {
```
TiagoFuelber, line 84: This method doesn't need to be async because is not returning the promise:

Could be like this
```
@Method()
  validateAndSubmit({ onSuccess }) {
    this.formController.validateAndSubmit(() => this.sendData(onSuccess));
  };
```

packages/webcomponents/src/components/business-form-stepped/business-form-stepped.tsx
colganmeanor: Good call!

packages/webcomponents/src/components/business-form-stepped/business-core-info/business-core-info-form-step.tsx
colganmeanor: Very nice to do this instead of binding stuff in the constructor. This removes a handful of lines from several files. 🙏🏻

TiagoFuelber (approved):
Awesome! Approved

PR #426, opened 2024-03-20
Identity form component

### LINKED ISSUE #425: Create identity form component
For the owners portion of the business form, we are planning to render an individual form component for each owner. 

Action: create an independent form component that hits the newly updated Identity endpoint to GET, PATCH, and POST identities associated with a given business.

colganmeanor:
### Create Identity Form Component for Business Owners

Links
-----

Closes #425 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------


packages/webcomponents/src/components/business-forms/owner-form/owner-form.tsx
```diff
@@ -0,0 +1,243 @@
...
+                />
+              </div>
+              <div class="col-12 col-md-6">
+                <form-control-number-masked
+                  name="phone"
+                  label="Phone Number"
+                  defaultValue={ownerDefaultValue?.phone}
+                  error={this.errors.phone}
+                  inputHandler={this.inputHandler}
+                  mask={PHONE_MASKS.US}
```
jakemer, line 175: Nice, I didn't realize we had a masking mechanism! I was looking at the schema file, and phone is only `required`. I know we're going to do some work on the validation requirements for this form later, but I wonder if we should make the validation schema for this field match whatever the mask is expecting... thoughts?

jakemer (approved):

packages/webcomponents/src/components/business-forms/owner-form/owner-form.tsx
colganmeanor: Good call! 

I went ahead and added some additional validation to the phone part of the schema for this component. We can add it to the forms that require phone numbers after the dust settles.

jakemer (approved):

Issue #424, opened 2024-03-18
All components emit expiredToken event


TiagoFuelber:
When there is a 401, check if the jwt is expired and emit an event for expiredToken

PR #427, opened 2024-03-21
Adds a expiredToken event to the PaymentDetails

TiagoFuelber:
This adds the `expiredToken` event to the `justifi-payment-details` component.
It's emitted when `payment-details-core` emits `error-event` with the error "Not Authenticated` which is also new.

Links
-----

https://github.com/justifi-tech/web-component-library/issues/424

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [ ] Add unit tests

Developer QA steps
--------------------
Expired token: `eyJraWQiOiJqdXN0aWZpLTQxNTczZmIyNjU0MDVmNDY0Y2UyIiwidHlwIjoiSldUIiwiYWxnIjoiUlMyNTYifQ.eyJpc3MiOiJodHRwczovL2F1dGguanVzdGlmaS1zdGFnaW5nLmNvbS8iLCJhenAiOiJ3Y3RfNVNKckZPWTI0YXJ5bUJRc2xTY3hvRiIsInN1YiI6IndjdF81U0pyRk9ZMjRhcnltQlFzbFNjeG9GQHNlc3Npb25zIiwiYXVkIjoiaHR0cHM6Ly9hcGkuanVzdGlmaS5haS92MSIsImd0eSI6ImNsaWVudC1jcmVkZW50aWFscyIsInBlcm1pc3Npb25zIjpbIndyaXRlOmFjY291bnQ6YWNjXzMzTzlESWdJZVMzOTFMSUxIRVIxZmYiLCJ3cml0ZTpidXNpbmVzczpiaXpfNm02VDlESUFtaW1WRlFZd2FoT2xQQiJdLCJleHAiOjE3MTA3OTc1MjksImlhdCI6MTcxMDc5MzkyOSwicGxhdGZvcm1fYWNjb3VudF9pZCI6ImFjY18zRklibDNUSWhUVUJoWGt3YVRYNTlaIn0.XTIyf8YsbqRKeSOPGj-LS-W2HC8D1QoXVOo7TMeRM6fE-TA2n1xc5naF1v57ddTvbg8MKqTBG0ssPsBvgk0db7WVtMNgM_WPSMje-0vRokJAH9aCyhYgKDA9IBhEAZ4ibbJiXMw11gwiB8Xi89952r-j01Qt-4kTVK2_pyeWvjnDdANM5w-nkgFdKffeIdc-C1P7Yvze3auiRZSR7PvGBJVgrO7U9L7T9EeyLer0xj26hlgIX-45AucSN3NVr7WZkjrX95ikC2fKhw5XG6U6yCD8MUFJ5wT4WO-yh4Ma7qdJg-ZK7gie1RiWF57ve4W8M3_uFOb8xrCz1AKqXjcugA`

[ ] - When trying to load a payment with an expired token, you should see in the actions tab on Storybook that an `expiredToken` event was logged.
[ ] - When loading a payment with a valid token, actions tab should be clear


.vscode/launch.json
```diff
@@ -3,22 +3,13 @@
     {
       "type": "node",
       "request": "launch",
-      "name": "Spec Test Current File",
-      "cwd": "${workspaceFolder}/stencil-library",
-      "program": "${workspaceFolder}/stencil-library/node_modules/.bin/stencil",
-      "args": ["test", "--spec", "--", "${fileBasename}"],
+      "name": "Debug Jest Tests",
```
TiagoFuelber, line 6: I think no one was using the old two configurations for local debugging and they were also outdated since the directory restructure.

packages/webcomponents/src/api/services/__mocks__/payment.service.ts
```diff
@@ -0,0 +1,13 @@
+import paymentDetail from '../../mockData/mockPaymentDetailSuccess.json';
```
TiagoFuelber, line 1: I found out recently that if we create a mock inside a folder named `__mocks__` we can use this mock on a test file like this: (this folder should be in the same level of the module we want to mock)

`jest.mock('../../services/payment.service.ts`)`

just adding this before the other imports in the test file. 👍

packages/webcomponents/src/components/payment-details/payment-details.tsx
```diff
@@ -60,17 +63,21 @@ export class PaymentDetails {
     }
   }
 
-  paymentService = new PaymentService();
+  handleError = (event) => {
+    if (event.detail === API_ERRORS.NOT_AUTHENTICATED) {
+      this.tokenExpired.emit({ bubbles: true, composed: true });
```
colganmeanor, line 68: I think we ought to emit something in this event - right now its set up to emit an object with the properties `bubbles: true, composed: true`

Perhaps it should simply emit the event, or the `event.detail`. 

Also, I think you would then move this information to line 42 - so that the event declaration reads as:
`@Event({ bubbles: true, composed: true }) tokenExpired: EventEmitter<any>`

colganmeanor (commented):
Nice work T! 

Everything looks really good, just left a question / comment about what we we are emitting with the event from `payment-details.tsx`

packages/webcomponents/src/components/payment-details/payment-details.tsx
TiagoFuelber: You're right. I added this to test something and forgot to remove. It's not necessary. 👍

colganmeanor: Cool cool - I saw your update. I do still wonder if we will eventually want to emit the `event.detail` or the `error` inside `tokenExpired.emit()` but I guess we can decide that whenever we want. 

Good work, thanks for the update! 

I'll approve for now, but still need to do Dev QA for this one

colganmeanor (approved):

PR #434, opened 2024-03-27
Implement new identity form into business form

### LINKED ISSUE #405: Add owners-form back to business-form-stepped
Now that business-form-stepped component is starting to stand on it's own two feet - we should begin adding back in the owners form sub component

Let's create `business-owners-form-step.tsx` or update it if it already exists. 

Like the other sub components, this component should fetch it's own data, instantiate its own Form Controller instance using the Owners schema, and then successfully Patch edits to individual Owners, or Post new owners that are being created via the form.

colganmeanor:
### Implement Owner Form component back into `business-form-stepped`

This PR re-implements the owners form into `business-form-stepped`. 

Some other small updates worth mentioning:

- Created Identity type file and abstracted identity / owner / representative things out of Business.ts and into Identity.ts
- Removed `private` from form.ts validate method so it could be used independently in owner-form

Links
-----

Closes #405 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------

We will want to test the owners section of the business form excessively. 
In order to test, you should be comfortable using insomnia to create businesses, patch existing businesses, create web component tokens, etc.

In the case that you are going through the form with a newly created business that has no owner data:

- [ ] The owners section of the form should render with a blank identity form. 
   - [ ] The identity form that is rendered should not make an API Call since there is no owner-id passed to it.
   - [ ] Attempting to submit without filling out all required fields for the current form should trigger validation and prevent submission
   - [ ] Clicking `Add` when filling out the initial form should call submit for just that particular form.
   - [

### DIFF
--- a/packages/webcomponents/src/api/Business.ts
+++ b/packages/webcomponents/src/api/Business.ts
@@ -1,3 +1,4 @@
+import { Identity } from './Identity';
 import { BankAccount } from './shared';
 
 export enum BusinessStructure {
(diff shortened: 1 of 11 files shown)


TiagoFuelber:
Excellent work @colganmeanor 🙌
I have just found one issue (an edge case) on QA (this is not super critical, it can be bypassed returning to the previous step). 

The case:
- Load a new business without owners
- On the owners form, add a new owner without filling up the empty one that shows, so we have 2 empty owners.
- Now try to delete one of them
- Both owners are deleted and the screen becomes not usable with no owner form and no "Add owner" button (I have to click previous and get back to step 5)

Also I would have some suggestions regarding the code but knowing that we have other tasks for refactoring and adding unit-tests it would make more sense to do it all together  👍

TiagoFuelber (approved):
Nice work! 💪 👨‍💻

colganmeanor:
@TiagoFuelber hey man, just pushed up some code that should tighten up the experience with the Remove Owners button showing or not showing. 

Ready to check again at your convenience!

TiagoFuelber (approved):
Code reviewed and Dev QA'd 👍 
Great job!

Issue #437, opened 2024-04-03
Investigate Storybook example mock data


jakemer:
Is it possible to mock API responses in Storybook so that the data doesn't change? This could be helpful for a handful of reasons:
- Chromatic visual diffing will not be an issue because of changing display data
- We wouldn't need to worry about auth tokens or rotating them
- We wouldn't need to worry about maintaining example businesses / payments data / etc
- For components such as our future Checkout component, which takes a session that only lasts an hour, it is extremely difficult to provide a working example

PR #440, opened 2024-04-08
Relocate mock data to root folder

TiagoFuelber:
This only relocates the JSON mock files that were initially inside `api` folder, to the root folder of the project, so they could be shared between both Stencil components and Storybook mock services.

Links
-----

#437 

Developer considerations
--------------

- On any task

Developer QA steps
--------------------
[ ] - There no specific test here, just ensure not component was broken, all unit tests still passes and build is not broken.


colganmeanor (approved):
Approved CR and Dev QA

🦧

PR #444, opened 2024-04-08
Mock all components

TiagoFuelber:
This adds all the routes to mock the services for all components in Storybook.

Links
-----
#437 

Developer considerations
--------------

- On any task

Developer QA steps
--------------------
[ ] - All components should be returning mock data

### DIFF
--- a/apps/docs/stories/utils/mockAllServices.ts
+++ b/apps/docs/stories/utils/mockAllServices.ts
@@ -1,16 +1,46 @@
 import { createServer } from 'miragejs';
 import mockBusinessDetails from '../../../../mockData/mockBusinessDetails.json';
+import mockGrossPaymentChart from '../../../../mockData/mockGrossVolumeReportSuccess.json';
+import mockPayment from '../../../../mockData/mockPaymentDetailSuccess.json';
+import mockPayments from '../../../../mockData/mockPaymentsSuccess.json';
+import mockPayout from '../../../../mockData/mockPayoutDetailsSuccess.json';
+import mockPayouts from '../../../../mockData/mockPayoutsSuccess.json';
 
 export const mockAllServices = () => {
   createServer({
     routes() {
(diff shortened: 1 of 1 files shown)


colganmeanor (approved):
Approved CR and Dev QA ✅

🥑

colganmeanor:
I approved this, but do want to make a note about Gross Payment Chart's mock data. 

We may need to add an optional prop to this component to allow us to provide a start/end date for the data. 

The way this component works is to look at `Today` in date/time and grab the last 30 days of payment gross data to fill out the chart. So its always looking at TODAY, and the columns on the chart should represent the 30 days that happened before TODAY. 

So, when we use mock data like this, it loads the chart just fine, but it is strange to look at the data and see that the data for the majority of the columns is several weeks or months behind the last column `TODAY`. 

The reason I'd suggest us to add an optional prop is so we can have the chart looking nice and consistent, and still use the mockdata that is here without havving to update it everytime. IDK - maybe there is a better or easier solution to just update the mock data dates every time the data is loaded? 🤔

TiagoFuelber:
> I approved this, but do want to make a note about Gross Payment Chart's mock data.
> 
> We may need to add an optional prop to this component to allow us to provide a start/end date for the data.
> 
> The way this component works is to look at `Today` in date/time and grab the last 30 days of payment gross data to fill out the chart. So its always looking at TODAY, and the columns on the chart should represent the 30 days that happened before TODAY.
> 
> So, when we use mock data like this, it loads the chart just fine, but it is strange to look at the data and see that the data for the majority of the columns is several weeks or months behind the last column `TODAY`.
> 
> The reason I'd suggest us to add an optional prop is so we can have the chart looking nice and consistent, and still use the mockdata that is here without havving to update it everytime. IDK - maybe there is a better or easier solution to just update the mock data dates every time the data is loaded? 🤔

Good observation!

Another option would be to use the proxyApi in Storybook, and if possible, only use mock data for this component for the Chromatic build, so we still would have screenshot tests.

PR #446, opened 2024-04-11
Fix Chromatic build

TiagoFuelber:
This fixes the Chromatic build command script and adds some delay to make room for the components rendering.

Links
-----
#437 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------

[ ] - A fresh new Chromatic build should be triggered and the screenshots should display the components with mock data. (For now, it will be broken because the base screenshots are still the ones in `main`. This will likely be fixed when this change get merged into `main`)


colganmeanor (approved):
Approved CR and QA

🦤

PR #466, opened 2024-04-19
BusinessFormStepped - Change state text input to select dropdown

colganmeanor:
### BusinessFormStepped - Change state text input to select dropdown

On step 2 of the BusinessFormStepped / PaymentProvisioning component - updated State input from `form-control-text` to `form-control-select`

- Added more secure validation
- moved `state-options.ts` into utils directory
- added helper function to help catch incoming data 
   - It's possible that someone could have created a Business with via API, or the older version of the form, and they typed in a full state name, such as `Alabama`. This helper function matches `Alabama` to the value `AL` when it loads data into the form and submits the address payload. 

Links
-----

Closes #456 

Developer considerations
--------------

- On any task
  - [ ] Previous unit and e2e tests are passing
  - [ ] Perform dev QA

Developer QA steps
--------------------

Step 2 of the form should now use `form-control-select` for `State`, instead of `form-control-text`.

For testing, please create a new business or patch an existing one via Insomnia. 

- [ ] Set the legal address state to something valid like `AL` - verify that this loads correctly in Step 2 and shows the appropriate label. 
   - [ ] Validation and submit fires correctly

- [ ] Set the legal address state to something that a user might have typed before, like `Alabama` - verify that this loads correctly in Step 2 and shows the appropriate label. 
   - [ ] Validation and submit fires correctly
   - [ ] After submitting this step, check the payload or response - the `state` value sent to the backend should now be `AL`

- [ ] Set the legal address state to something invalid like `France` - verify that nothing shows in the state dropdown menu, its value should be undefined. 
   - [ ] Attempt to validate - verify that validation fires correctly and triggers error state for State input. 
   - [ ] Change it to something valid via the dropdown menu
   - [ ] Validation and submit fires correctly

PR #475, opened 2024-04-25
PaymentProvisioning - Improve Address Form Validation and Logic

### LINKED ISSUE #461: PaymentProvisioning - Improve validation for Addresses
Address validation needs some improvements

**Postal Codes:**
- `e` is allowed to be inputted for some reason
- Limit input to 5 characters - currently you can put any number of numerical characters
- [https://docs.google.com/spreadsheets/d/1KpgaPMl0X9OWxm7D3yAoBDOqkHcpTtWInZLS0OxqPSw/edit#gid=503272030&range=24:24](https://docs.google.com/spreadsheets/d/1KpgaPMl0X9OWxm7D3yAoBDOqkHcpTtWInZLS0OxqPSw/edit#gid=503272030&range=24:24)

**Street Address:**
- Unsure what this should be exactly, but you can currently enter a single character and its considered valid
- [https://docs.google.com/spreadsheets/d/1KpgaPMl0X9OWxm7D3yAoBDOqkHcpTtWInZLS0OxqPSw/edit#gid=503272030&range=20:20](https://docs.google.com/spreadsheets/d/1KpgaPMl0X9OWxm7D3yAoBDOqkHcpTtWInZLS0OxqPSw/edit#gid=503272030&range=20:20)

### LINKED ISSUE #470: BusinessFormStepped: country for representative missing
**Is your feature request related to a problem? Please describe.**
Representative Step:
There is no `Country` field in the form and the form doesn't seem to autofill that information in the post request.
However, we require the representative country when the business gets provisioned.

Screen Shot 2024-04-23 at 2 05 03 PM

To fix: either provide country field in form or autofill.

colganmeanor:
### Minor Improvements and Consolidations to Address Forms

- Consolidates near-duplicate component directories
   - `business-address-form` and `business-address-form-stepped` are now a single directory, renamed `identity-address-form` since this component is only used in Rep and Owner form components
- Add `country` field to `identity-address-form` component. Show country on Representative and Owner Forms
   - `country` field is disabled, and auto-fills with `USA / United States` based on class logic
   - this change also applies to the Representative section of the original `BusinessForm`
- Speaking of the original business-form, Updated `state` text field in `legal-address-form.tsx`, `form-control-text` to select menu, `form-control-select`
   - this mirrors the change we added to `PaymentProvisioning` before we released the beta. (PR - #466 )
- Added improved logic and validation for Address Fields 
   - Some fields such as line1, line2, and city have min and max character limits, and disallow special characters
   - Postal code field has been converted to `form-control-text` and added some logic to prevent alphabetical and special characters, as well as limit length. 
- Consolidated address schemas now that the `legal-address-form` and `identity-address-form` have the exact same fields. 

Links
-----

Closes #470
Closes #461 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------

- [ ] `country` input (form-control-select) now appears on `Representative` and `Owner` forms
  - [ ] should be disabled, and auto-fills with `USA` just as it does in the Business Address section
- [ ] In the original `BusinessForm` - the business address section should now show `state` as a select menu just as it does in `PaymentProvisioning`

Some validation QA steps:
Note: these can be tested on either `BusinessForm` or `PaymentProvisioning`. Both components are using the same schema info, so these fields will all get the same validation rules. 

- [x] Legal Address / Line 1 AND Line 2 - should have min 5, max 100 characters - no special characters allowed
- [x] City - min 2 characters, max 100 characters - no special characters allowed
- [x] Postal code - must be 5 characters exactly - should only allow numerical values
- [x] New address validation rules should work the same with `allowOptionalFields` set as `True` OR `False`. If `True`, then the fields should have the same validation rules, but should be nullable if you remove the value.


TiagoFuelber:
Hey @colganmeanor @jakemer 
Just wanna confirm, do we want to make the `address line 2` have a minimum of 5 characters?
image

TiagoFuelber:
The city seems to be limiting characteres to 50 and not 100
image

colganmeanor:
> Hey @colganmeanor @jakemer Just wanna confirm, do we want to make the `address line 2` have a minimum of 5 characters? image


I did this on purpose, but I'm willing to reverse it for Line 2. I basically just applied the same validation rules as Line 1 - but yeah I'm imagining maybe people would just put a number, or a single letter here or something. Hard to predict. 

I could add a specific rule for Line 2 that has no min / max, but still checks for special characters. How does that sound?

packages/webcomponents/src/components/business-forms/owner-form/identity-address/identity-address-form.tsx
```diff
@@ -66,16 +64,29 @@ export class BusinessAddressForm {
...
           <div class="col-12 col-md-6">
-            <form-control-number
+            <form-control-text
               name="postal_code"
               label="Postal Code"
               defaultValue={this.defaultValues?.postal_code}
               error={this.errors?.postal_code}
+              maxLength={5}
+              keyDownHandler={(e) => filterPostalInput(e)}
               inputHandler={(name, value) => this.inputHandler(name, value)}
```
TiagoFuelber, line 75: Just a small improvement: updating the assignment of the inputHandler in your component to use a direct reference to the method, like this:

```
inputHandler={this.inputHandler}
```

Benefits of this approach include:


Error Traceability: Using a named function improves the clarity of stack traces during debugging. It's easier to identify issues because the function name appears directly in error messages.

Performance Optimization: By referencing the handler directly, the function is not recreated on each render. This prevents unnecessary re-renders of the form-control-text component.

Code Readability and Inspection: While using direct method references might seem less explicit about what arguments are passed, most development environments allow you to hover over the method name to see its signature, thus maintaining an easy way to understand the expected arguments.

colganmeanor: Great call - thank you!

packages/webcomponents/src/components/business-forms/owner-form/identity-address/test/identity-address.e2e.ts
```diff
@@ -0,0 +1,11 @@
+import { newE2EPage } from '@stencil/core/testing';
```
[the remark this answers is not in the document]
colganmeanor: I think I agree - it doesn't seem like we're doing anything with these at the moment. 

I don't think I'll do that in this PR, but we should make an issue for this and put it on the board to discuss. 👍🏻

packages/webcomponents/src/components/business-forms/owner-form/identity-address/identity-address-form.tsx
```diff
@@ -22,6 +24,10 @@ export class BusinessAddressForm {
     this.address = { ...this.address };
   }
 
+  constructor() {
+    this.inputHandler = this.inputHandler.bind(this);
```
TiagoFuelber, line 28: We can set `inputHandle` as an arrow function to get `this` bound automatically.

e.g
```
inputHandler = (name: string, value: string) => {
...
}
```

TiagoFuelber:
> > Hey @colganmeanor @jakemer Just wanna confirm, do we want to make the `address line 2` have a minimum of 5 characters? image
> 
> I did this on purpose, but I'm willing to reverse it for Line 2. I basically just applied the same validation rules as Line 1 - but yeah I'm imagining maybe people would just put a number, or a single letter here or something. Hard to predict.
> 
> I could add a specific rule for Line 2 that has no min / max, but still checks for special characters. How does that sound? @TiagoFuelber

Yes, it sounds good! I think it's reasonable for this line. Thanks

TiagoFuelber:
> > The city seems to be limiting characteres to 50 and not 100 image
> 
> @TiagoFuelber - I purposefully set it to max 50 characters - seemed reasonable to me.
> 
> Are there cities with longer names than that? 🤔
> 
> It looks like the text in this screenshot is the names of several cities?

😆  The city in the screenshot seems to be the name locals call Bangkok which is considered the longest city name in the world. 

But I only asked that because in the QA steps is says it should limit to 100. But I am also ok with 50. There must be not much city names bigger than 50 🤔 😆

TiagoFuelber (approved):

packages/webcomponents/src/components/business-forms/owner-form/identity-address/identity-address-form.tsx
colganmeanor: Good call - just updated that one! Thank you!

Issue #484, opened 2024-05-02
PaymentForm not applying custom styles on the styled example


TiagoFuelber:
Payment Form is not applying the custom styles on the styled example.

This is happening only for the PaymentForm.

1. Go to https://storybook.justifi.ai/?path=/story/payment-facilitation-payments-payment-form--styled
2. See error

**Expected behavior**
The form should have the custom styles applied

**Screenshots**
![image](https://github.com/justifi-tech/web-component-library/assets/22480988/52daadc8-f852-4d72-b860-fd3c764ea63a)

 -There's no error on the screen neither on browser console

PR #486, opened 2024-05-03
Fix decorator

TiagoFuelber:
This fixes the custom styles that were not being correctly applied.
It also fixes a Typescript error that was happening in this file for some time.

Links
-----
#484 

image

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

- On bugs
  - [ ] Chromatic should have this issue covered

Developer QA steps
--------------------
[ ] - http://localhost:6006/?path=/story/payment-facilitation-payments-payment-form--styled Should render the Payment Form with the custom styles applied
[ ] - (Forcing an error) - Fill the form with a ZIP code longer then 5 digits. Submit it. An error event should be logged in the actions tab.


colganmeanor (approved):
Approved CR and QA 

🍩

PR #487, opened 2024-05-03
PaymentForm disableSubmitButton method

jakemer:
We have a method for `enableSubmitButton` - this method is used to re-enable the PaymentForm submit button in a implementers submit flow. Because the PaymentForm only handles tokenization of payment methods, the implementer then needs to use the token to make a payment and tell the form to re-enable the submit button when that is done.

We are now adding a `disableSubmitButton` to give implementers more control, such as disabling payments prior to a prerequisite to payment 

Links
-----

Developer considerations
--------------

- On any task
  - [x] Perform dev QA
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
- Pull the branch/install/build/run dev server
- In storybook, go to the PaymentForm example
- In the developer tools / console, run `const el = document.querySelector('justifi-payment-form')`, then `await el.disableSubmitButton()`
- [x] The submit button should now be disabled

### DIFF
--- a/.changeset/mean-falcons-shop.md
+++ b/.changeset/mean-falcons-shop.md
@@ -1,0 +1,6 @@
+---
+"@justifi/webcomponents": patch
+"@repo/docs": patch
+---
+
+Added disableSubmitButton method to PaymentForm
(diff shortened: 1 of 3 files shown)


TiagoFuelber (approved):
Approved!

PR #488, opened 2024-05-06
Create form-control-date input to manage Birthdays for Owners and Representatives

### LINKED ISSUE #459: PaymentProvisioning - Improve DOB Inputs
For the DOB Month fields (visible in Rep and Owner Forms), user simply enters a number input. 

We should consider changing this to a dropdown menu with Month names as visible options, values sent to backend are numbers though

[https://docs.google.com/spreadsheets/d/1KpgaPMl0X9OWxm7D3yAoBDOqkHcpTtWInZLS0OxqPSw/edit#gid=503272030&range=12:12](https://docs.google.com/spreadsheets/d/1KpgaPMl0X9OWxm7D3yAoBDOqkHcpTtWInZLS0OxqPSw/edit#gid=503272030&range=12:12)

Additionally - there seems to be a bug related to year input: 
Entering 2 digits  in the year input is allowed, but the backend interprets that, or automaticlly updates something like `84` to `1900`. 
Feels like a backend question, but worth considering if we can do anything to prevent weird behavior on the frontend
[https://docs.google.com/spreadsheets/d/1KpgaPMl0X9OWxm7D3yAoBDOqkHcpTtWInZLS0OxqPSw/edit#gid=503272030&range=2:2](https://docs.google.com/spreadsheets/d/1KpgaPMl0X9OWxm7D3yAoBDOqkHcpTtWInZLS0OxqPSw/edit#gid=503272030&range=2:2)

colganmeanor:
### Create form-control-date input to manage Birthdays for Owners and Representatives

This PR updates the business-form and payment-provisioning components by adding a single input to manage date of birth, replacing the 3 separate inputs we previously had for day, month, and year. 

While implementing, I had to adjust the existing layout since we removed 2 inputs. DOB and SSN are now on the same line when screens are wide enough. 

Styling note: I chose to hide the calendar icon on the input for now - which removes the ability to click on the input and show the calendar picker part of the input. For right now, I wanted to leave that out so we can decide how to style it eventually. 

Links
-----

Closes #459 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------

- [ ] Complete BusinessForm using new date input - try to break it plz
- [ ] Complete PaymentProvisioning using new date input - try to break it plz


TiagoFuelber:
@colganmeanor 
I don't know if we should consider this a bug or not. But the field allows inputting an year with more than 4 digits.
WDYT?
image

colganmeanor:
> @colganmeanor I don't know if we should consider this a bug or not. But the field allows inputting an year with more than 4 digits. WDYT? image

Oh that is interesting... I hadn't tried that. 

Looking online, this seems to be a common complaint about the native html date input. :/ 

I'm seeing we could try to use javascript to prevent this behavior, but I'm not sure if there is a perfect solution. I was just trying out something like: 

```
handleFormControlInput = (event: any) => {
    const target = event.target;
    const name = target.getAttribute('name');
    const value = target.value;

    // Limit the year to 4 digits
    const dateParts = value.split('-');
    if (dateParts.length > 0 && dateParts[0].length > 4) {
      dateParts[0] = dateParts[0].substr(0, 4);
      target.value = dateParts.join('-');
    }

    this.inputHandler(name, target.value);
    this.formControlInput.emit(target.value);
  }
```

But its not perfect, it makes sure that year is only 4 digits, but if you keep typing in the input it can just overwrite the 4 digits that are there with whatever next characters you type. 

We do have validation in place for the input, that will never allow this.

```
export const dobValidation = (role: string) => {
  return (
    string()
    .test('min', 'Enter a valid date', (value) => {
      const date = new Date(value);
      const minDate = new Date('1902-01-01');
      return date >= minDate;
    })
    .test('age', `${role} must be at least 18 years old`, (value) => {
      const date = new Date(value);
      const minAgeDate = new Date();
      minAgeDate.setFullYear(minAgeDate.getFullYear() - 18);
      return date <= minAgeDate;
    })
    .transform(transformEmptyString)
  )
};
```

Even if using the `allow-optional-fields` prop a user should not be able to submit bad data.

TiagoFuelber (approved):
Looks good! Nice work!

Issue #485, opened 2024-05-02
Add Authorization info to PaymentForm and BankAccountForm docs


Claudia108:
**Is your feature request related to a problem? Please describe.**
- Confirm the PaymentForm works with a web component token
- Add an Authorization section to PaymentForm and BankAccountForm similar to the section in the CardForm component.
- Update the web component token part of the Authorization section to: 
```
These tokens are generated by your backend services using the [Web-Component API](https://docs.justifi.tech/api-spec#tag/Web-Component-Tokens/operation/CreateWebComponentToken). Each token can be scoped to perform a set number of actions and is active for 60 minutes. 
When creating a web component token for this specific component you'll need to use the role `write:tokenize:account_id` (use the account id you also pass as a prop).
```
(Once we confirm that the PaymentForm works with the web component token I will add it to the list of roles)

- Add the specific web component token role needed for each web component. The complete list is here: https://docs.justifi.tech/infrastructure/webComponentTokens#roles-need-for-each-component

Add this below the props section of other web components:

```
Authorization

Authorization is performed by passing a `web-component-token` as auth-token.

A `web-component-token` is generated by your backend services using the [Web Component Tokens API](https://docs.justifi.tech/api-spec#tag/Web-Component-Tokens/operation/CreateWebComponentToken). Each token can be scoped to perform a set number of actions and is active for 60 minutes.
When creating a web-component-token for this specific component you'll need to use the role `<role for component>` (use the account id you also pass as a prop separately). or for business related components (use the business id you also pass as a prop separately)
```

PR #494, opened 2024-05-09
add web-component-token prop and authorization docs

TiagoFuelber:
**Description**
This PR introduces a new property auth-token to the PaymentForm component to enhance clarity and reduce confusion. Previously, the `client-id` property was overloaded to accept a token, which could be misleading due to its name. Now, `auth-token` will exclusively handle this, aligning its functionality more intuitively with its name.

Additionally, this PR updates the Authorization section across all component documentation to reflect these changes and ensure consistency.

Links
-----
#485 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [ ] Add unit tests
  - [ ] Add documentation

Developer QA steps
--------------------
**PaymentForm**
* First, get an `authToken` at `{{ _.capital_url }}/oauth/token`

 - [x] If both `clientId` and `authToken` are missing, an error event should be emitted.
 - [ ] PaymentForm should work if either `authToken` and `clientId`.
 - [ ] If `clientId` is used, `Authorization` header should have the `Basic` prefix on the `tokenize` request.
 - [ ] If `authToken` is used, `Authorization` header should have the `Bearer` prefix on the `tokenize` request.

**Docs**

 - [ ] All components should display an "Authorization" section with the correct instructions according with #485


Claudia108:
Looks so much better!
Here are a few things I have noticed:
- BusinessDetails: only show role `read:business:business_id`
- GrossPaymentChart: role should be `read:account:account_id`
- PaymentsList: Could you please move the Authorization section above the event usage example?
- BankAccountForm: code example:
- - Could you please add the web component token option to he code example like this
`bankAccountForm.tokenize('CLIENT_ID/WEB-COMPONENT-TOKEN', paymentMethodData, 'ACCOUNT_ID')`
-- Could you please add this comment above that instead of the current sub account comment:
```
/*
        * ACCOUNT_ID is the sub account for which you are tokenizing
        * WEB-COMPONENT-TOKEN can be used in place of CLIENT_ID.  WEB-COMPONENT-TOKENs are 
        * created using JusitFi's API as described above.
      */
```
- CardForm code example: In the comment above the `.tokenize` method can you please use `sub account` instead of `seller account`?
- PaymentForm: 
-- props section: could you please move the `web-component-token` props right below the `client-id`?
-- code example: could you please add the `web-component-token` there as a comment and update the props comment above?

``` 
<justifi-payment-form
    client-id="test_df97f04afebc3c018de30df3562d7cdd"  // web-component-token="your-web-component-token"
    account-id="acc_5Et9iXrSSAZR2KSouQGAWi"
    bank-account="true"
    card="true"
    email="test@email.com"
    submit-button-text="Something you want here"
 ></justifi-payment-form>```
```

```
/*
* Props definition:
*
* client-id OR web-component-token use either for authorization (see Authorization section above)
* account-id is the merchant sub account that you intend to tokenize the payment on
*/
```

TiagoFuelber:
@Claudia108  Suggested: 

"I also have some additional suggestion and am wondering if I should create a separate story for it or if you'd rather add them to this pr:
I think it would be great to add pre requisites for each component in the description (right below the title of the component). For example the BusinessForm description could say something like:
Component to render a business onboarding form. You will need to first create a business via API (link to api endpoint) to get the business-id required for this component.
2. I'd love to see a bit more detail on the description for some of the props. e.g. auth-token could be: " Auth token string , is the web-component-token described below" or business-id  "Business ID string , the id of the business created via API."

apps/docs/stories/components/business-details-docs.stories.mdx
```diff
@@ -1,36 +1,52 @@
...
-<Meta
-  title="Entities/BusinessDetails"
-  Components="justifi-business-details"
-/>
+<Meta title="Entities/BusinessDetails" Components="justifi-business-details" />
 
 <Title />
-Component to render detailed information about a business.
+Component to render detailed information about a business. You will need to first
+create a business via [API](https://docs.justifi.tech/api-spec#tag/Business) to get
```
Claudia108, line 10: Could you please change `[Api]`  to `[Business API]`?

apps/docs/stories/components/business-form-docs.stories.mdx
```diff
@@ -1,14 +1,21 @@
...
-<Meta
-  title="Entities/BusinessForm"
-  Components="justifi-business-form"
-/>
+<Meta title="Entities/BusinessForm" Components="justifi-business-form" />
 
 <Title />
-Component to render a business onboarding form.
+Component to render a business onboarding form. You will need to first create a business
+via [API](https://docs.justifi.tech/api-spec#tag/Business) to get the business-id
```
Claudia108, line 17: Same here

apps/docs/stories/components/payment-provisioning-docs.stories.mdx
```diff
@@ -1,14 +1,24 @@
...
 
 <Meta
   title="Entities/Payment Provisioning (BETA)"
   Components="justifi-payment-provisioning"
 />
 
 <Title />
-Component to render a business onboarding form, segmented into multiple steps.
+Component to render a business onboarding form, segmented into multiple steps. You
+will need to first create a business via [API](https://docs.justifi.tech/api-spec#tag/Business)
```
Claudia108, line 20: Same here

Claudia108 (approved):

apps/docs/stories/components/payment-form.stories.tsx
```diff
@@ -133,16 +113,22 @@ const meta: Meta = {
 }
 
 const Template = (args: any) => {
+  // Set the prop only if it is defined
+  const getAttributeString = (attrName: string) => {
+    return args[attrName] ? `${attrName}="${args[attrName]}"` : '';
```
jakemer, line 118: Nice 👍 Definitely gets tiring writing over and over and I don't think we've been entirely consistent so this would help as a shared utility function

jakemer (approved):

apps/docs/stories/utils/documentation-components.tsx
```diff
@@ -49,3 +49,89 @@ export const SummaryElement = (props: { title: string, children: any }) => {
...
+
+  return (
+    <>{getCodeForRole(actions[0])} or {getCodeForRole(actions[1])}</>
+  );
+};
+
+const AuthTokenDescription = () => (
+  <>
+    <strong>Auth Token</strong >: These tokens are generated by your backend services using the {' '}
+    <a href="https://docs.justifi.tech/apiFundamentals/authentication">
```
Claudia108, line 72: Could you please instead link to `https://docs.justifi.tech/api-spec#tag/API-Credentials/operation/CreateAccessToken` and call it `Create Access Token API`?

apps/docs/stories/utils/base-args.ts
```diff
@@ -14,7 +14,7 @@ const args: ArgValues = {
 const argTypes: ArgTypes = {
   'auth-token': {
     type: 'string',
-    description: 'Auth token `string`',
+    description: 'Auth token `string`, the web component token you create via API. See Authorization below',
```
Claudia108, line 17: Is this generic and will also show up for payment, card and bank account form? In that case we would have to show an alternative auth token text.
I'd say let's just say `'Auth token `string`, see Authorization below' which would cover every use case.

apps/docs/stories/components/checkout-docs.stories.mdx
```diff
@@ -1,11 +1,22 @@
...
+<Meta
+  title="Payment Facilitation/Payments/Checkout"
+  Components="justifi-checkout"
+/>
 
 <Title />
-Component to render the necessary fields to enter proper credit card information.
-
+Component to render the necessary fields to enter proper payment method information
+and effectuate payment.
```
Claudia108, line 19: Can we say `process a payment` instead of `effectuate payment`?

PR #519, opened 2024-05-16
Create form control file input

### LINKED ISSUE #507: Document upload component (internal)
Create a document upload component that is an internal component for our library, to be used in other components.

We should attempt first to use a native `file` input with extra styling as necessary (does bootstrap have a class for this?), before trying out any libraries.

colganmeanor:
### Create `form-control-file` input 

This PR adds a basic form control component that will be used for file uploading in the business forms. There is no submission logic in this PR since this is just about creating the base component and corresponding test coverage. Submission logic will be handled when we implement this into an existing form. 

Links
-----

Closes #507 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [X] Add unit tests

Developer QA steps
--------------------

- [ ] Verify that component can render - use storybook or generic HTML page serve
- [ ] Verify that component can open file browser and accept file uploads


packages/webcomponents/src/components/form/form-control-file.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+@Component({
+  tag: 'form-control-file',
+  styleUrl: 'form-control-file.scss',
+  shadow: true,
+})
+export class FileInput {
+  @Element() el: HTMLElement;
+  @Prop() label: string;
+  @Prop() name: any;
```
TiagoFuelber, line 20: we probably want  `error`, `disabled` and maybe `label` to be optional

TiagoFuelber (approved):
Just a tiny comment, but it's approved! Nice test suite!

packages/webcomponents/src/components/form/form-control-file.tsx
colganmeanor: @TiagoFuelber 

Hmmm, I'm not sure that we need to do that since its written the same way in each of the other form-control components. 

I wonder if that's something we should fix in each of the form-control components, though I'm not entirely sure why we have never gotten warnings about this before. 🤔

PR #524, opened 2024-05-24
Refact(Api): Change Api props to receive an object instead of two arguments

TiagoFuelber:
This PR updates the Api class to accept a single object parameter instead of two separate arguments. The new usage is `Api({ authToken, iframeOrigin })` instead of `Api(authToken, iframeOrigin)`.

The reason being that the analytics endpoint does not receive an authToken. Consequently, the Api class has been updated to handle cases where `authToken` is absent.

This changes nothing in the components usage.

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------
[ ] - All components should be still loading and saving data without errors

### DIFF
--- a/packages/webcomponents/src/api/Api.ts
+++ b/packages/webcomponents/src/api/Api.ts
@@ -22,16 +22,33 @@ export interface IApiResponseCollection<T> extends IApiResponse<T> {
   page_info: PagingInfo;
 }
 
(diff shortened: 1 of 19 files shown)


jakemer (approved):
Nice, changes make sense and makes our Api functions more expandable!

Issue #496, opened 2024-05-09
Analytics: send version and device info on initialization


TiagoFuelber:
We should have an `/analytics` endpoint in the web component proxy so that we can report on component usage and errors where it makes sense. We may decide to capture analytics in other cases down the road, so that is something to keep in mind with the approach and pattern for this.

On `init` we would make a request to the `/analytics` endpoint with a payload something like the following:
```
{
  event_type: 'init',
  data: {
    component_name: 'Checkout',
    component_version: '4.11.0',
    client_user_agent: 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:124.0) Gecko/20100101 Firefox/124.0',
    client_platform: 'MacIntel',
    client_origin: 'https://somewebsite.com'
  }
}
```

On `error`, it would be something like:
```
{
  event_type: 'error',
   data: {
    component_name: 'Checkout',
    component_version: '4.11.0',
    client_user_agent: 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:124.0) Gecko/20100101 Firefox/124.0',
    client_platform: 'MacIntel',
    client_origin: 'https://somewebsite.com',
    error: // whole javascript error object if possible
  }
}
```

PR #529, opened 2024-05-27
Feat: Implement analytics in all components

TiagoFuelber:
This PR is for initialization of Analytics class in all components.

In the test file of each component is necessary to mock the `JustifiAnalytics` class to avoid an error related to browser environment in node. And also because we don't want to send analytics in a test environment.

Links
-----
#496 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------
[ ] - Each component in Storybook should send Analytics data on init and on other events.


packages/webcomponents/src/components/business-forms/payment-provisioning/payment-provisioning.tsx
```diff
@@ -23,7 +24,24 @@ export class PaymentProvisioning {
   @State() errorMessage: string = '';
   @State() currentStep: number = 0;
   @State() totalSteps: number = 5;
-  @Event({eventName: 'click-event'}) clickEvent: EventEmitter<BusinessFormClickEvent>;
+  @Event({ eventName: 'click-event' }) clickEvent: EventEmitter<BusinessFormClickEvent>;
+
+  analtyics: JustifiAnalytics;
```
colganmeanor, line 29: `analytics` is misspelled in this file - but its correct in every other component file! 

Just need to update this from `analtyics` => `analytics`

packages/webcomponents/src/components/payments-list/payments-list.tsx
```diff
@@ -3,6 +3,7 @@ import { PaymentService } from '../../api/services/payment.service';
 import { makeGetPayments } from './get-payments';
 import { ErrorState } from '../details/utils';
 import { ComponentError, ComponentErrorCodes, ComponentErrorSeverity } from '../../api/ComponentError';
+import JustifiAnalytics from '../../api/Analytics';
```
colganmeanor, line 7: Not sure where is best to leave this comment - based on the description of the PR I am expecting every event emitted by a component to trigger a post to the `analytics` endpoint. 

I am triggering the `clickEvent` for clicking a payment row - the event is captured and logged in storybook Actions, but there is no analytics post when this event occurs. Is this intentional? Or should this be adjusted?

packages/webcomponents/src/components/business-forms/business-form/business-form.tsx
```diff
@@ -29,6 +30,24 @@ export class BusinessForm {
   @Event({ eventName: 'click-event' }) clickEventNew: EventEmitter<BusinessFormClickEvent>;
   @Event({ eventName: 'clickEvent' }) clickEventOld: EventEmitter<BusinessFormClickEvent>;
```
colganmeanor, line 32: In `business-form` - I am correctly seeing an analytics post for `submitted` events, but I am not seeing an analytics post for `clickEvents` in this component.

packages/webcomponents/src/components/business-forms/payment-provisioning/payment-provisioning.tsx
```diff
@@ -23,7 +24,24 @@ export class PaymentProvisioning {
   @State() errorMessage: string = '';
   @State() currentStep: number = 0;
   @State() totalSteps: number = 5;
-  @Event({eventName: 'click-event'}) clickEvent: EventEmitter<BusinessFormClickEvent>;
+  @Event({ eventName: 'click-event' }) clickEvent: EventEmitter<BusinessFormClickEvent>;
```
colganmeanor, line 27: Not really relevant to this PR, but I'm realizing after looking at your code that have `submitted` documented as an event that gets logged on the Payment Provisioning form, but we aren't emitting it at all. I need to fix that in a separate PR. 🛠️

packages/webcomponents/src/components/payments-list/payments-list.tsx
colganmeanor: Same applies to the click event for `payouts-list` as well

TiagoFuelber: I am not sure either what would be the ideal behavior on this.

Based on the current logic, the Analytics class will only track the events emitted by the component where the Analytics was initiated. Hence the click-event is emitted by a lower level component, the Analytics is not aware of that.

This was the easiest implementation but we absolutely can change this logic with a little bit more code if we decide to track all events down the tree. WDYT @jakemer @colganmeanor

colganmeanor: I think we can resolve this thread - we discussed this morning and agreed that the click events dont matter, at least right now for this PR.

packages/webcomponents/src/components/business-forms/business-form/business-form.tsx
TiagoFuelber: @colganmeanor @jakemer 
Decided to use a fixed list of events we wanna track with Analytics instead of trying to get them dynamically.

https://github.com/justifi-tech/web-component-library/pull/529/files#diff-b38ac37a9b31c34f4641f64952f06108d22b0e903b20f59ee82b921dff2ec425

packages/webcomponents/src/api/Analytics.ts
```diff
@@ -54,7 +48,7 @@ class JustifiAnalytics {
   };
 
   private trackCustomEvents() {
-    this.eventEmitters = getEventNames(this.componentInstance);
+    this.eventEmitters = ['submitted', 'error-event'];
```
TiagoFuelber, line 51: Instead of trying to get the event-names from each component, it's actually easier to use a fixed list here for now. We can wait until we feel the need for a more complex approach.

colganmeanor (approved):

Issue #527, opened 2024-05-24
PaymentForm: bank-account prop not working


Claudia108:
**Describe the bug**
In the PaymentForm when you set the `card` prop to `false` and the `bank-account` prop to `true`, the form loads the card form instead of the bank account form.

**To Reproduce**
Go to the PaymentForm storybooks example and set `card` prop to `false` and `bank-account` prop to `true`.
The form loads the credit card form.

**Expected behavior**
The form should only load the bank-account form

**Screenshots**
<img width="917" alt="Screen Shot 2024-05-24 at 3 12 36 PM" src="https://github.com/justifi-tech/web-component-library/assets/15485124/e102a182-84d2-443b-8773-4702c8ab16c2">

PR #530, opened 2024-05-28
Add getSelectedPaymentMethodType method to PaymentForm

jakemer:
Fix issue with PaymentForm where card form is being shown instead of bank account form when card prop is false and bank-account prop is true

Links
-----
Fixes https://github.com/justifi-tech/web-component-library/issues/527

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
Run the branch locally, and in the PaymentForm example try the following
- set card prop to `false` and bank account to `false`, the card form should show (something needs to show, card is the fallback)
- set card prop to `true` and bank account to `true`, the payment method toggle should show and user can flip between card form and bank account form
- card form `true`, bank account `false`
- bank account `true`, card `false

### DIFF
--- a/packages/webcomponents/src/components/payment-form/payment-form.tsx
+++ b/packages/webcomponents/src/components/payment-form/payment-form.tsx
@@ -49,6 +49,22 @@ export class PaymentForm {
     this.submitButtonEnabled = false;
   }
 
+  showPaymentMethodTypeSelector() {
+    return this.card && this.bankAccount;
+  }
+
+  getSelectedPaymentMethodType() {
+    if (this.showPaymentMethodTypeSelector()) {
+      return this.selectedPaymentMethodType;
+    } else if (this.card) {
+      return PaymentMethodTypes.card;
+    } else if (this.bankAccount) {
+      return PaymentMethodTypes.bankAccount;
+    } else {
+      return PaymentMethodTypes.card;
+    }
+  }
+
   paymentMethodSelectedHandler(event: CustomEvent) {
     const paymentMethodType: PaymentMethodTypes = event.detail;
     this.selectedPaymentMethodType = paymentMethodType;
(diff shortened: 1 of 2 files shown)


Claudia108 (approved):
Looks good to me!

packages/webcomponents/src/components/payment-form/payment-form.tsx
```diff
@@ -49,6 +49,22 @@ export class PaymentForm {
     this.submitButtonEnabled = false;
   }
 
+  showPaymentMethodTypeSelector() {
```
colganmeanor, line 52: Everything works correctly - one tiny suggestion: 

If you rewrite these as `getter` methods then they don't need to be written as functions to be called. You could do a very quick change:

```
  get PaymentMethodTypeSelector() {
    return this.card && this.bankAccount;
  }

  get SelectedPaymentMethodType() {
    if (this.PaymentMethodTypeSelector) {
      return this.selectedPaymentMethodType;
    } else if (this.card) {
      return PaymentMethodTypes.card;
    } else if (this.bankAccount) {
      return PaymentMethodTypes.bankAccount;
    } else {
      return PaymentMethodTypes.card;
    }
  }
```

And then when you are referencing this later you can simply reference it as `this.PaymentMethodTypeSelector` or `this.SelectedPaymentMethodType`

jakemer: Very nice suggestion however it leads to a naming conflict with a `@State` variable and resolving that will likely require a `setter`, so I am going to get this out as-is and we can always tweak that later (although I think we will be deprecating `PaymentForm` in the not-so-distant future)

PR #538, opened 2024-05-30
Fix PaymentForm prop logic for card and bankAccount

### LINKED ISSUE #537: PaymentForm - card / bankAccount prop bug
There is currently a bug on production with the `PaymentForm` component - prop specification for `card` or `bankAccount` is not working as intended. 

Setting `bankAccount = true` and `card = false` does not show the Bank Account Form as expected. Props should work as described

- If both props are True - show the card form with the Payment Method Selector
- If both props are false - show the default - card form
- If one prop is true, and the other is false - show the one that is true without the selector

colganmeanor:
### Fix PaymentForm prop logic for card and bankAccount

Fixes an existing bug on the PaymentForm related to the card and bankAccount props. Removes the default `true` for card prop that never gets changed, and thus never allows the other prop to work correctly. 

Links
-----

Closes #537 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------

### DIFF
--- a/.changeset/weak-hairs-notice.md
+++ b/.changeset/weak-hairs-notice.md
@@ -2,4 +2,4 @@
 "@justifi/webcomponents": patch
 ---
 
-- Fully resolved issue with the `PaymentForm` component props that show bank-account form.
+- Fully resolved issue with the `PaymentForm` component props that show the bank-account form.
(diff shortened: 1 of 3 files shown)


jakemer (approved):
Nice and simple! Good work

Claudia108 (approved):

Claudia108 (approved):
Even better

PR #546, opened 2024-06-04
Payment provisioning document uploading

### LINKED ISSUE #521: PaymentProvisioning - Document Upload Step
Image

Add new step to PaymentProvisioning form - Document Upload

- Step will go after Owners Form
- Should have 4 fields: 
  - `voided check`
  - `balance statement`
    - 2022 and 2023?
  - `bank statement`

colganmeanor:
### Add document upload form step to `PaymentProvisioning`

This PR adds document uploading functionality to the stepped business form / payment provisioning component. 

Additionally, it abstracts `Document` classes, interfaces, types to Document.ts
Makes some changes to `form-control-file` input 

Links
-----

Closes #521 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------

New form step added to PaymentProvisioning form - document uploading - last step
Each file uploaded should have an appropriate `document_type`, which is determined by the file input that you use to upload a particular file

I'm sharing a folder with some sample files to use for testing - see slack thread where I post this PR for review. 👀 

- [x] Verify form validation fires if user attempts to click `Submit` without interacting with the file inputs
- [x] Verify that form step works as expected when `allow-optional-fields === true` (Form validation should not fire, submit button should emit a click event
- [x] Verify that document uploading works - should see 2 requests made in the dev console network tab for each document you select. The first request should be to the documents endpoint to create a document record, and the other request for this document will be the Upload to the s3 bucket. 
- [ ] Verify that you are able to upload multiple documents from a single file input, and all will be correctly uploaded with the expected `document_type`

It will be best to refresh the page and check the response from the request to view the business data - look for the `documents` property to confirm again successful uploads. Documents fully uploaded will have a `uploaded` status. 

If you are working with a fresh business, and using this form step - you should only see documents with `uploaded` status.


packages/webcomponents/src/api/Document.ts
```diff
@@ -0,0 +1,133 @@
...
+  presigned_url: string | null;
+  status: string;
+  updated_at: string;
+}
+
+export class Document implements IDocument {
+  public business_id: string;
+  public created_at: string;
+  public description: string | null;
+  public document_type: string;
```
TiagoFuelber, line 22: I think this could be

```suggestion
  public document_type: EntityDocumentType;
```

packages/webcomponents/src/api/Document.ts
```diff
@@ -0,0 +1,133 @@
...
+  public file_name: string;
+  public file_type: string;
+  public id: string;
+  public identity_id: string;
+  public metadata: any;
+  public platform_account_id: string;
+  public presigned_url: string | null;
+  public status: string;
```
TiagoFuelber, line 30: And here:

```suggestion
  public status: EntityDocumentStatus;
```

packages/webcomponents/src/components/business-forms/payment-provisioning/document-uploads/business-document-upload-form-step.tsx
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  storeFiles = (e: CustomEvent<FileSelectEvent>) => {
+    const fileList = Array.from(e.detail.fileList) as File[];
+    const docType = e.detail.document_type;
+    const documentList = fileList.map(file => new EntityDocument({ file, document_type: docType }, this.businessId));
+    this.documentData[docType] = documentList;
+  }
+  
+  sendData = async (onSuccess?: () => void) => {
```
TiagoFuelber, line 128: Should this also receive an `onError` callback to be called on line 130 and 134?

packages/webcomponents/src/api/Document.ts
colganmeanor: Hmmm good catch! 

I actually wonder if we should just remove that entire class from the code for now - the `Document` class that's declared in the top of the file is not used anywhere else - that code used to live in the Business.ts file but even then was never used anywhere. 

The `IDocument` interface can stay, since `Business.ts` imports it and uses it to type its `document` property. 

What do you think? Keep or remove this class and its constructor? 

@TiagoFuelber

packages/webcomponents/src/components/business-forms/payment-provisioning/document-uploads/business-document-upload-form-step.tsx
```diff
@@ -0,0 +1,236 @@
...
+      method: 'PUT',
+      body: docData.fileString,
+    })
+
+    return this.handleUploadResponse(response);
+  }
+
+  handleUploadResponse = (response: any) => {
+    if (response.error) {
+      this.serverError.emit({ data: response.error, message: DocumentFormServerErrors.sendData });
```
TiagoFuelber, line 113: We could either rename this event to `error-event` or add `server-error` to Analytics events list.
`packages/webcomponents/src/api/Analytics.ts:63`

packages/webcomponents/src/components/business-forms/payment-provisioning/document-uploads/business-document-upload-form-step.tsx
```diff
@@ -0,0 +1,236 @@
...
+export class BusinessDocumentFormStep {
+  @Prop() authToken: string;
+  @Prop() businessId: string;
+  @Prop() allowOptionalFields?: boolean;
+  @State() formController: FormController;
+  @State() errors: any = {};
+  @State() documents: any = [];
+  @State() documentData: EntityDocumentStorage = new EntityDocumentStorage();
+  @Event({ bubbles: true }) submitted: EventEmitter<BusinessFormSubmitEvent>;
+  @Event() formLoading: EventEmitter<boolean>;
```
TiagoFuelber, line 24: Beside keeping event names consistent, another good reason for using kebab-case on our event names is that I think it will be easier to track events automatically on Analytics in the future. 

```suggestion
  @Event({ eventName: 'form-loading' }) formLoading: EventEmitter<boolean>;
```

packages/webcomponents/src/components/business-forms/payment-provisioning/document-uploads/business-document-upload-form-step.tsx
```diff
@@ -0,0 +1,236 @@
...
+  @Event() serverError: EventEmitter<DocumentFormServerErrorEvent>;
```
TiagoFuelber, line 25:
```suggestion
  @Event({ eventName: 'server-error' }) serverError: EventEmitter<DocumentFormServerErrorEvent>;
``` 

or

```suggestion
  @Event({ eventName: 'error-event' }) serverError: EventEmitter<DocumentFormServerErrorEvent>;
```

packages/webcomponents/src/api/Document.ts
TiagoFuelber: I didn't noticed it was not being used. In that case I think it's alright to remove the class. This will keep our code cleaner. Good catch!

packages/webcomponents/src/components/business-forms/payment-provisioning/document-uploads/business-document-upload-form-step.tsx
colganmeanor: hey @TiagoFuelber  - I completely agree with your suggestions about the way we are handling the error event. 

I want to implement these changes in a PR that will follow this one - the way it is currently set up for the PaymentProvisioning component is that the parent component listens for `serverError` to be emitted from its child components (the different form steps) and then uses that to display an error on the page. 

I want to move in this direction and get the `error-event` emitted from the component at each step, the way you've outlined here, but to do so I will also need to make some refactors to all the other form step components. 

I think its a little too much to jam into this PR - so I'd like to do this work in another PR that I will go ahead and start working on 👍🏻

packages/webcomponents/src/components/business-forms/payment-provisioning/document-uploads/business-document-upload-form-step.tsx
colganmeanor: In line with my other response - yes I totally agree about passing an `onError` callback to this function. I want to update this in this component, and the other form step components in another PR that will address error handling logic for the entire PaymentProvisioning component

Issue #551, opened 2024-06-11
Not possible to import components using ES modules


TiagoFuelber:
A client has reported this issue:

When trying to use a component like `import { JustifiBusinessForm } from '@justifi/webcomponents';` an error occurs: `Module '"@justifi/webcomponents"' has no exported member 'JustifiBusinessForm'`

Also, the docs still says it's possible to install `@justifi/react-components` https://storybook.justifi.ai/?path=/docs/introduction--docs#react

PR #556, opened 2024-06-14
add docs section for importing components using ES modules

TiagoFuelber:
This PR adds to the docs some instructions on how to install the package from NPM and how to import component modules in a ES module way.

This also removes the instructions about using `@justifi/react-components` in the intro docs, hence this React package is no longer supported/recommended.

Links
-----
#551 

Developer considerations
-------------

Developer QA steps
--------------------
The only thing that changed here was the docs. So nothing will break. Anyways, the new instructions can be tested in any local demo-app


colganmeanor (approved):
Looks good to me - thanks for cleaning this up! 

I am approving - but will wait and see if @Claudia108 has any notes.

Claudia108 (approved):
Looks good to me! Thank you for the updates.

PR #570, opened 2024-06-24
Create form help text and form error text components

### LINKED ISSUE #569: Abstract form help text logic to its own component
Currently we do not have a unified system to implementing help or error text to our form-control components. 

Let's abstract this logic from our various form-control components, and create a `form-help-text` component that can conditionally render help or error text if provided.

colganmeanor:
### Create form help text and form error text components

This PR helps clean up the `form-control` components by abstracting existing logic for help and error text and handling logic in 2 new components that will be conditionally rendered inside each `form-control` component: 

- `<form-control-help-text />`
- `<form-control-error-text />`

For this PR, I only implemented new components into the `form-control-file` component, since this is currently the only form control component that is utilizing both `helpText` and `errorText`.

An additional PR will follow this that will complete implementation for the other form control components. 

Additionally, I removed some scss files that really were not being used since the component was getting all the styles it needed from its parents higher up the tree. 

Links
-----

Closes #569 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------

Everything should look the same as it was before. 

The new `<form-help-text />` component is only implemented to the `document-upload` form step, for the `form-control-file` inputs. 

- [x] When `helpText` is supplied to `form-control-file`, then you should see `<small class='form-text text-muted'>{text}</small>`
- [x] When `errorText` is supplied to `form-control-file`, then you should see `<small class='form-text text-danger'>{text}</small>`

Just load the document upload form step, and click `submit` without touching any of the inputs. Error state should be triggered for some of the fields which will show the proper error text from the form controller, and the correct `text-danger` class.


packages/webcomponents/src/components/form/form-helpers/form-help-text/form-help-text.tsx
```diff
@@ -0,0 +1,34 @@
...
+    return this.errorText ? 'text-danger' : 'text-muted';
+  }
+
+  render() {
+
+    if (!this.helpText && !this.errorText) {
+      return null;
+    }
+
+    if (this.errorText) {
```
jakemer, line 22: Would it be better to keep help text and error text separate? We may want the flexibility of displaying both? I think with the way this is we are coupling the two when the purposes are possibly too different

colganmeanor: Agreed, and done. Good call. 👍🏻

packages/webcomponents/src/components/form/form-helpers/form-control-help-text/form-control-help-text.tsx
```diff
@@ -0,0 +1,21 @@
...
+  @Prop() helpText: string;
+
+  render() {
+
+    if (!this.helpText) {
+      return null;
+    }
+
+    return (
+      <small class='form-text text-muted' id='formHelpText'>
```
jakemer, line 16: I think this is going to result in some issues of having duplicate ids on the page. Can we make it unique somehow?

colganmeanor: Thanks for pointing that out! 

I updated this to accept `name` as a prop, which it can get from its parent form control input, and it now interpolates the `name` into the `id` of the help or error text component. 👍🏻 

Marking this comment as resolved.

jakemer (approved):
Nice work 👍

PR #585, opened 2024-07-01
Remove unused style sheets business forms

### LINKED ISSUE #586: Remove shadow-dom and un-needed scss files from BusinessForm and PaymentProvisioning children components
We are learning how best to style and generate stencil components as we proceed with development, upon reflection I see that we do not need several scss files that were created for while I was developing the `PaymentProvisioning` components. Giving each of these components a scss file, and a shadow dom just makes styling more complicated.

The top level file in a component's DOM is the one that primarily needs to have a shadow dom, the component's children components don't necessarily need one, and don't necessarily need their own scss files if they share so many styles with the parent.

Action: Remove scss files that aren't used or needed for children components of `BusinessForm` and `PaymentProvisioning`

Action: Add shadow dom to `PaymentProvisioning` and `BusinessForm`, and remove shadow dom from their children components where its not necessary.

colganmeanor:
### Remove unused style sheets from BusinessForm and PaymentProvisioning children components

We are learning how best to style and generate stencil components as we proceed with development, upon reflection I see that we do not need several scss files that were created for while I was developing the PaymentProvisioning components. Giving each of these components a scss file, and a shadow dom just makes styling more complicated. 

The top level file in a component's DOM is the one that primarily needs to have a shadow dom, the component's children components don't necessarily need one, and don't necessarily need their own scss files if they share so many styles with the parent. 

This PR does the following: 

- Adds `shadow: true` to the `BusinessForm` and `PaymentProvisioning` components
- Removes `shadow: true` and the corresponding scss file from most of the children components rendered in `BusinessFrom` and `PaymentProvisioning`
- unrelated - makes a small adjustment to the Owners form step by giving the entire form step it's own `legend` text, and adjusting the size and weight of the `legend` text that renders on each individual `owner-form` that is rendered at this step. 

Links
-----

Closes #586 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------

Nothing should be changed on the user experience here. 

- [x] Ensure tests are passing. 
- [x] Please build this branch locally, and take a good look at both `BusinessForm`, and `PaymentProvisioning` components. Everything should look the way it did before, since we didn't remove any active styles, just duplicate code and files. 
- [x] The only thing that should actually change a little, is that there should now be a `legend` / title text for the overall Owners Form Step in `PaymentProvisioning`. You should see a legend in the same position as every other form step, and each of the Owner Forms rendered here have had their `legend` / title text adjusted to a smaller font size, and semi-bold font weight.


TiagoFuelber:
I noticed that the fields on Business Legal address have a smaller margin top of 8px then the fields on other sub-forms that have a 16px margin top.

image

colganmeanor:
> Not sure if this was already an issue before but I noticed that the fields on Business Legal address have a smaller margin top of 8px then the fields on other sub-forms that have a 16px margin top.
> 
> image

@TiagoFuelber ah good catch! I checked on prod storybook, and this is how it has always been. It seems like it just comes down to a different `row` class from bootstrap. 

Both the `legal_address` form components, in BusinessForm, and PaymentProvisioning have now been updated to use `row gy-3` in line with the other form section components. 👍🏻

PR #588, opened 2024-07-03
Documents on file component

### LINKED ISSUE #587: PaymentProvisioning - Show uploaded documents on Document Form Step
Users are inquiring about, and requesting information on the Document Upload Form step - requesting some communication on this form step about which documents have already been uploaded. 

After discussing with Claudia, we have landed on the following plan:

- Create simple table component (perhaps named: `documents-on-file` that is conditionally rendered in the form step. 
- table renders with 3 columns: `file_name`, `file_type`, `document_type`
- table cells should not have borders
- Should table rows or columns have borders? 🤔 
- Section Header: `Documents on File`

Additionally we want to adjust the validation for this form step to be flexible for repeat form submissions. This might be a separate PR, but the logic should be:

> IF a document exists in the list of documents with the corresponding document type, then we will NOT require that field to be completed again.

colganmeanor:
### Add new table component to display on Document Upload Form Step in `PaymentProvisioning`

This PR is a response to customer feedback about being able to see existing / already uploaded documents on the `PaymentProvisioning` component's document form step. 

- Added a new component to conditionally render in the document form step: `documents-on-file`
- Refactored render logic in document form step - no longer takes in `paymentVolume` as a prop. Gets it directly from the business API call it performs on render, and captures existing documents at the same time
- Changed form controller initialization logic slightly in this component, just needed to initialize it at the correct time after data has come back from the business endpoint. 
- Adjusted schema validation logic for document upload form step - now takes into account existing documents that business has already uploaded. 
  - NOTE: at the time, this is fairly simple logic. It just checks that there is an existing document with the corresponding `document_type`. This is about all we can do at this time, until we have more data or properties to work with from the backend. 

Links
-----

Closes #587 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------

As with most DEV QA tasks for this component, I'd recommend editing `PaymentProvisioning` to only render the document upload form step, and perhaps the terms and conditions form step to ensure that validation, submission, and nextStep are all working correctly. 

- [x] If loading the form with fresh business data, should only see the document inputs, table will not be rendered
- [x] If loading the form with business data that has already uploaded documents, table will be rendered. Verify that the documents shown in the table are accurate to the documents returned in the business API response (check the network tab and preview the data)

For the next QA step - we need to ensure validation is still working consistently with the old behavior, and taking existing documents into account now. 

Create a new business and set `business_payment_volume` to a value that is `249,000` or less.
- [x] Of the 4 fields that render - only `voided_check` should be required. Try to submit without interacting with the inputs, and verify that validation still fires correctly. 
- [x] Submit a voided_check document here, and then refresh the form. Come back to this form step. Verify that you may now proceed to the next form step without validation issues. 

Create a new business and set `business_payment_volume` to a value that is `250,000` or higher.
- [x] Of the 5 fields that render - only `voided_check` and `bank_statement` should be required. Try to submit without interacting with the inputs, and verify that validation still fires correctly. 
- [x] Submit the required documents, and then refresh the form. Come back to this form step. Verify that you may now proceed to the next form step without validation issues. 

Create a new business and set `business_payment_volume` to a value that is `1,000,000` or higher.
- [x] Of the 7 fields that render -`voided_check`, `bank_statement`, `balance_sheets`, and `profit_loss_statements` should be required. Try to submit without interacting with the inputs, and verify that validation still fires correctly. 
- [x] Submit the required documents, and then refresh the form. Come back to this form step. Verify that you may now proceed to the next form step without validation issues. 
- [x] One more step here - reload this form step once more. Even though you can proceed without uploading documents here, just try to upload any additional documents (required or not required). Verify that any documents you selected are properly uploaded once you click `Next`


packages/webcomponents/src/components/business-forms/schemas/schema-validations.ts
```diff
@@ -232,37 +233,51 @@ export const routingNumberValidation = string()
...
-    then: (schema) => schema.required('Please select one or more files'),
-    otherwise: (schema) => schema.nullable(),
-  });
-}
\ No newline at end of file
+  let existingDoc = documents.some((doc) => doc.document_type === EntityDocumentType.balanceSheet);
+
+  if (existingDoc || allowOptionalFields) {
+    return documentUploadValidation.nullable();
+  } else if (vol >= 1000000 && !allowOptionalFields) {
```
TiagoFuelber, line 269: I'm thinking here about maybe putting this number in a variable explaining what it is. For instance:
`const paymentVolumeHigherLimit = 1000000`

WDYT?

TiagoFuelber (approved):
Great work! Approved with a minor comment. ✅

packages/webcomponents/src/components/business-forms/schemas/schema-validations.ts
colganmeanor: Totally agreed. Good call! 

Just pushed up this change :)

Issue #525, opened 2024-05-24
Storybook theme switcher and decorator


jakemer:
For the Checkout component storybook example we have a new decorator and "theme" arg that we want to use across all stories. This issue is about abstracting the theme concept but not about applying it to every story, that can happen in subsequent issues.

For this issue let's:
- abstract the ["theme" arg/argType](https://github.com/justifi-tech/web-component-library/blob/main/apps/docs/stories/components/checkout.stories.tsx#L21) to our `StoryBaseArgs` utility
- abstract [the new decorator which applies the theme stylesheets](https://github.com/justifi-tech/web-component-library/blob/main/apps/docs/stories/components/checkout.stories.tsx#L65)

- Reference: https://storybook.js.org/docs/essentials/toolbars-and-globals?ref=storybookblog.ghost.io

PR #589, opened 2024-07-08
Add 'custom-styled' prop to baseArgs

TiagoFuelber:
**Description**

This adds the `theme` prop to base-args in order to be possible to switch between "basic" and "custom" styles on the story controls.

As a consequence of this change, the "custom" extra stories for each component were removed. Now we have only the "basic" story with a "theme" prop under "theming" section on the story controls.

* In the prop description, added a link to the [Introduction](https://storybook.justifi.ai/?path=/docs/introduction--docs#styling-components-with-variables) to show an example of the CSS variables available.

image

(top toolbar buttons) 
* ~~The top toolbar buttons are default and it seems that they cannot be removed.~~ Yes they got removed 👍 

Links
-----
#525 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------

[ ] - Styleable components should now display just 'basic' story with a `theme` prop to switch the styles.

Components: 
- Payments List
- Payouts List
- Bank Account Form
- Card Form
- Payment Form
- Checkout ('light' and 'dark')

### DIFF
--- a/apps/docs/.storybook/main.ts
+++ b/apps/docs/.storybook/main.ts
@@ -1,18 +1,26 @@
-import { mergeConfig } from "vite";
-import react from "@vitejs/plugin-react";
+import { mergeConfig } from 'vite';
+import react from '@vitejs/plugin-react';
 
 const config = {
-  framework: "@storybook/web-components-vite",
-  stories: ["../stories/**/*.stories.@(mdx|ts|tsx)"],
+  framework: '@storybook/web-components-vite',
+  stories: ['../stories/**/*.stories.@(mdx|ts|tsx)'],
   addons: [
-    "@storybook/addon-links",
-    "@storybook/addon-essentials",
-    "@storybook/addon-docs",
+    '@storybook/addon-links',
+    {
+      name: '@storybook/addon-essentials',
+      options: {
+        measure: false,
+        outline: false,
+        backgrounds: false,
+        viewport: false,
+      },
+    },
+    '@storybook/addon-docs',
   ],
   docs: {
     autodocs: true,
   },
-  staticDirs: ["../public"],
+  staticDirs: ['../public'],
   async viteFinal(config) {
     const storybookMocksEnabled = process.env.VITE_STORYBOOK_MOCKS_ENABLED;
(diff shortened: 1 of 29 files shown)


colganmeanor (commented):
Hey @TiagoFuelber - great job with this PR. Really awesome to see Storybook cleaned up here! :D 

A few questions / comments:

Screenshot 2024-07-08 at 11 08 50 AM

- Is it possible to have the `styles` controls in the control table minimized / collapsed by default? When the storybook example loads, it takes up a lot of vertical space. 
- Could we give the section with the `styles` controls a section header the same way we have for `Props` and `Events` below it? 
- Is it possible to have the `styles` control section shown ONLY if the `custom-styled` prop is set to True? 🤔 
- Now that we have removed the `Styled` components from the left nav menu, can we rename the examples from `Basic` to something more appropriate, perhaps `Example`? 🤔

TiagoFuelber:
> Hey @TiagoFuelber - great job with this PR. Really awesome to see Storybook cleaned up here! :D
> 
> A few questions / comments:
> 
> Screenshot 2024-07-08 at 11 08 50 AM
> 
> * Is it possible to have the `styles` controls in the control table minimized / collapsed by default? When the storybook example loads, it takes up a lot of vertical space.
> * Could we give the section with the `styles` controls a section header the same way we have for `Props` and `Events` below it?
> * Is it possible to have the `styles` control section shown ONLY if the `custom-styled` prop is set to True? 🤔
> * Now that we have removed the `Styled` components from the left nav menu, can we rename the examples from `Basic` to something more appropriate, perhaps `Example`? 🤔

Very good suggestions 🙌

Applied all except for collapsing the styles object, which actually is not possible on Storybook configuration. But rendering it conditionally seems to attenuate the problem.

jakemer:
This is looking good but I have a few thoughts:
- We should call the control "theme" and use the light/dark themes that are being used on the checkout component story - that'll require a multi-select for the control I suppose
- We should possibly hide the values if they can't be collapsed. I do see value in consumers being able to see the values right inline though vs somewhere else but that does bring me to my second point:
- We should add to the "styling" section of the docs and link to the themes so that consumers can see what a full stylesheet looks like and copy from it as desired

colganmeanor (approved):
🎨

Issue #579, opened 2024-06-27
Upgrade Turbo to v2


jakemer:
Also, check how to have a nice tests output running from root folder

PR #591, opened 2024-07-09
Migrate turbo to latest

TiagoFuelber:
This adds the changes needed for migrating Turbo from v1 to v2.
https://turbo.build/blog/turbo-2-0

Turbo v2 changes include improving the terminal output, which also solves the problem with the unit tests output not highlighted.

Links
-----

#579 
Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------
[ ] - Run `pnpm install` on root folder and try to build the project. It should build without errors.
[ ] - Run `pnpm test:watch` on the root folder. It should run the tests and the output should have syntax highlight

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "private": true,
+  "name": "web-component-library",
   "scripts": {
     "build": "turbo run build",
     "build:mocks": "turbo run build:mocks",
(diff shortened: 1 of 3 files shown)


colganmeanor (approved):
Approved ✅

Glad we got Turbo updated properly - thanks Tiago! 

That said, I tried running `pnpm test:watch` from the root directory on `main`, and on this branch, and I don't really feel like I noticed much of a difference between the two. 🤔 

It feels a little cleaner in the terminal to run tests with the updated turbo version though, so yeah it's definitely still a change we should approve and merge into main regardless.

PR #603, opened 2024-07-16
Post Provisioning Request in PaymentProvisioning component

### LINKED ISSUE #598: Make Provisioning API call in PaymentProvisioning
No extra UI component for this. 

In final form step, need to make an additional API call to the provisioning endpoint:
[https://docs.justifi.tech/api-spec#tag/Provisioning](https://docs.justifi.tech/api-spec#tag/Provisioning)

Use `business.legal_name` for `new_account_name` in request payload.

colganmeanor:
### Post Provisioning Request in PaymentProvisioning component

This PR adds logic to make a post request to the provisioning endpoint at the very end of the `PaymentProvisioning` form component. 

The following changes are committed by this PR: 

- Add logic to parent PaymentProvisioning component to fetch business and check if business has already submitted a provisioning request. 
  - In the case that it has submitted a provisioning request: the form buttons will be disabled, and an `error-event` will emit detailing this information. 
- Add logic to submit data to provisioning endpoint, makes post request after TermsAndConditions form step is successfully completed. 
- Emits event, `submitted`, which is emitted upon successful or failed provision request. 

Links
-----

Closes #598 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------

To test this, it will be best to use an existing business that has not been provisioned, or to create a new business. 

Complete the form with existing, un-provisioned business, or new business. Once terms and conditions form step is completed:

- [x] Verify that API request is made to provisioning endpoint with businessID and product: 'payment'
- [x] Verify that `submitted` event is emitted with the response from the request. 

Load the form again with your business that has been successfully provisioned:

- [x] Verify that form buttons are disabled - users may not use the form to adjust business info once provision request has been created
- [x] Verify that `error-event` is emitted with `provision-already-requested` as the error code.


packages/webcomponents/src/components/business-forms/payment-provisioning/payment-provisioning.tsx
```diff
@@ -17,51 +21,112 @@ import { ComponentError, ComponentErrorCodes, ComponentErrorSeverity } from '../
...
       });
-      return;
     }
-
-    this.refs = [this.coreInfoRef, this.legalAddressRef, this.additionalQuestionsRef, this.representativeRef, this.ownersRef, this.bankAccountRef, this.documentUploadRef, this.termsRef];
   }
 
-  disconnectedCallback() {
-    this.analytics.cleanup();
+  fetchBusiness() {
```
TiagoFuelber, line 76: Maybe I would rename this function to something like `setBusinessProvisioned` which feels more accurate with what the function is doing.

TiagoFuelber:
Nice work! 💯 

Another quick thing, I noticed that some readmes are updated when building this branch. Just need to check them in. 👍

TiagoFuelber (approved):
💯 approved!! Thanks man!

PR #604, opened 2024-07-17
Refactor PaymentProvisioning Parent Component

### LINKED ISSUE #349: Refactor component render logic: PaymentProvisioning
For `PaymentProvisioning`

- Create a service.ts to handle provisioning request logic.
- Create actions file to prepare API requests handled in `PaymentProvisioning` parent component

colganmeanor:
### Refactor `PaymentProvisioning` Parent Component

~This PR builds on work completed in #603.~
~#603 should be merged into main before this PR is reviewed~

- Refactors `PaymentProvisioning` to use core component structure. 
  - Creates `payment-provisioning-core` component which is now rendered in `payment-provisioning`
  - abstracts `ComponentStepMapping` to its own component to reduce clutter
  - abstracts form buttons to their own component to reduce clutter. 
- Removes `removeTitle` prop that was used to hide the form title. (the same effect can be achieved by passing an empty string to the `formTitle` prop instead. 

Links
-----

Closes #349 

Developer considerations
--------------

- On any task
  - [x] Perform dev QA

Developer QA steps
--------------------

QA for this PR will involve completing the PaymentProvisioning form. Ideally, nothing should have changed functionally with the form! It's probably best to test it with a new business freshly created via API. Though it would certainly be valuable to test with existing business data as well. 

That said, this PR removes the prop `removeTitle` in favor of simply using `formTitle` to remove the title if desired. A blank string can be passed to `formTitle` to achieve the same effect with less code on our end. 

- [x] Previous tests are passing. 
- [x] Verify that `formTitle` prop can be used to change the form title from the default `Business Information` to either something new, or removed via passing `''` as the prop value. 
- [x] Verify that each form step can be successfully completed as expected. 
- [x] Verify that error-event is emitted, and data is not fetched if auth token or business ID are not provided
- [x] Verify that provisioning api call happens at the end of the form
- [x] If a business has already requested provisioning, verify that the form emits an `error-event` on load stating that the business has already requested provisioning, and the button should be disabled. 
- [x] Verify that step counter is accurately showing the current number of steps, and the total number of steps
- [x] Verify that form validation fires correctly on each form step


apps/docs/stories/components/payment-provisioning.stories.tsx
```diff
@@ -18,7 +18,7 @@ const meta: Meta = {
     ...storyBaseArgs.argTypes,
     'form-title': {
       type: 'string',
-      description: 'This prop updates the value of the title displayed at the top of the form.',
+      description: 'This prop updates the value of the title displayed at the top of the form. Pass an empty string to remove the title.',
```
TiagoFuelber, line 21: Good idea!

TiagoFuelber (approved):
Awesome work dude! Approved

Issue #594, opened 2024-07-09
Error in console caused by Analytics


TiagoFuelber:
Analytics is causing an error trying to add an eventlistener to an element that is undefined

It's happening in the demo app

PR #607, opened 2024-07-18
Add an error bound

TiagoFuelber:
After some debugging it seems that the error os only caused because at some point the Analytics is called before the component is instantiated. I attempted to call Analytics in other life cycles but the error continued happening. 
Adding the validation to only call `addEventListener` when it's actually a function fixed the error without breaking the Analytics functioning

**Note: This bug was only reproducible on a demo-app**

* This adds an error bound to prevent addEventListener to be called on undefined

Links
-----
#594 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------

[ ] - Run a demo app (any) and add a symlink to the web-components. Any error regarding `addEventListener` is not a function should be visible on the console. Also a network request to Analytics should have happened on the network tab

### DIFF
--- a/.changeset/beige-lions-draw.md
+++ b/.changeset/beige-lions-draw.md
@@ -1,0 +1,5 @@
+---
+"@justifi/webcomponents": patch
+---
+
+Fix a console error caused by Analytics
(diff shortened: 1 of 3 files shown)


colganmeanor (dismissed):

colganmeanor (approved):
🛹

Issue #609, opened 2024-07-22
Storybook 8 migration


TiagoFuelber:
MIgrate from Storybook 7 to 8.

https://storybook.js.org/docs/migration-guide

Steps: 
- Run `pnpm dlx storybook@latest upgrade`
- Convert stories in MDX to CSF `pnpm dlx storybook@latest migrate mdx-to-csf --glob "src/**/*.stories.mdx"`
- Run codemod to migrate from CSF 2 to CSF3 `pnpm dlx storybook@latest migrate csf-2-to-3 --glob="**/*.stories.tsx" --parser=tsx`
- Add vite config:  https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#framework-specific-vite-plugins-have-to-be-explicitly-added

PR #610, opened 2024-07-22
Migration to Storybook 8

TiagoFuelber:
**Description**

[Storybook 8 changelog](https://storybook.js.org/releases/8.0)

This PR makes all changes necessary to migrate to Storybook 8.

* Followed this guide https://storybook.js.org/docs/migration-guide

Most changes where performed by codemods provided by Storybook.

I used this PR as an opportunity to restructure the `apps/docs/stories/components` folder to create a dedicated folder for each component

image
 

Links
-----

#609 

Developer considerations
--------------

- On any task

Developer QA steps
--------------------
- [x] Install all dependencies locally
- [x] Run the app a go over all stories to check everything is working as before


apps/docs/stories/components/CardForm/index.stories.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+import "@justifi/webcomponents/dist/module/justifi-card-form";
+
+const storyBaseArgs = new StoryBaseArgs([
+  "account-id",
+  "client-id",
+  "iframe-origin",
+  "theme",
+]);
+storyBaseArgs.argTypes["client-id"].table.disable = true;
```
jakemer, line 17: Is this being hidden in favor of web component tokens?

apps/docs/stories/components/BankAccountForm/index.stories.tsx
```diff
@@ -0,0 +1,155 @@
...
+import "@justifi/webcomponents/dist/module/justifi-bank-account-form";
+
+const storyBaseArgs = new StoryBaseArgs([
+  "account-id",
+  "client-id",
+  "iframe-origin",
+  "theme",
+]);
+storyBaseArgs.argTypes["client-id"].table.disable = true;
+storyBaseArgs.argTypes["account-id"].table.disable = true;
```
jakemer, line 18: Why are these being hidden? If it's because we only want to show how to use the component with web component tokens, should that be added in as an arg and displayed on the table?

packages/webcomponents/package.json
```diff
@@ -2,9 +2,10 @@
   "name": "@justifi/webcomponents",
   "version": "4.15.0",
   "description": "JustiFi Web Components",
-  "main": "dist/module/index.js",
+  "collection": "dist/collection/collection-manifest.json",
```
jakemer, line 5: Nice cleanup on this 👍

apps/docs/.storybook/main.ts
```diff
@@ -15,26 +15,35 @@ const config = {
         viewport: false,
       },
     },
-    '@storybook/addon-docs',
+    '@chromatic-com/storybook',
   ],
   docs: {
-    autodocs: true,
+    autodocs: false,
```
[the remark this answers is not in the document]
TiagoFuelber: Hmmm, I turned this to false in an attempt to solve an issue that was 2 docs visible for all components in the same time. But after reviewed this again, it looks like this setting was not the cause of the issue and the new folder structure seems to have solved the problem. 
I'm turning this to `true` again. Good catch 👍

apps/docs/stories/components/CardForm/index.stories.tsx
TiagoFuelber: Actually, this code was already there before in main. Not sure why either.

apps/docs/stories/components/BankAccountForm/index.stories.tsx
TiagoFuelber: These ones as well were not modified by this PR. I am not sure why they are being hidden. We can change that if we want it makes sense.

jakemer (approved):

PR #630, opened 2024-08-02
Checkout: Method to fill billing information

jakemer:
Just like we've done for `PaymentForm`, we need to expose a method that can be used to pre-fill the billing form: https://github.com/justifi-tech/web-component-library/blob/main/packages/webcomponents/src/components/payment-form/payment-form.tsx#L57

Customers use this when they already have the end-user's billing information on file or when it was collected in a previous step before checking out

Links
-----

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
- Check out the `fill-billing-form-test` branch of https://github.com/justifi-tech/checkout-demo
- Run the checkout demo using `npm run start` and open the browser to `localhost:3000`
- [x] With network tab open, click the "Fill Billing Address" button, and the form should fill. Fill in the rest of the form so that a payment can be submitted. Check the request payload to see that the address is correct in the payload
- [x] Repeat the last step, but before submitting change the values in the billing form and check the request payload to see that the address is correct

### DIFF
--- a/.changeset/chatty-trainers-repeat.md
+++ b/.changeset/chatty-trainers-repeat.md
@@ -1,0 +1,5 @@
+---
+"@justifi/webcomponents": minor
+---
+
+Add fillBillingAddress method to Checkout component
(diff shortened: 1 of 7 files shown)


TiagoFuelber (approved):
Nice work! Approved

PR #633, opened 2024-08-09
Checkout save new payment methods

### LINKED ISSUE #629: Checkout: Save payment method
For new payment methods (Card or Bank Account):
- If checkout has a `payment_method_group_id`, show a checkbox labeled "Save payment method"
- If the checkbox is checked, pass the `payment_method_group_id` along with the tokenize `paymentMethodMetadata` (Make sure that PaymentsJS application passes this along appropriately to the backend). See the tokenize call in the `PaymentMethodForm` component.

See [Create Payment Method](https://docs.justifi.tech/api-spec#tag/Payment-Methods) docs and expand `payment_method` section

colganmeanor:
### Add logic to save new payment methods

NOTE: this PR will need to be coordinated with the following PR for payments-js:
[Payments-js 136](https://github.com/justifi-tech/payments-js/pull/136
)

- Adds a checkbox below the new payment method billing form that will only render IF the checkout has an associated `payment_method_group_id`
- When checked / true, the checkout will submit the payment_method_group_id along with payment's metadata

Links
-----

Closes #629 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------
- Make sure that payment_method_group_id is being sent
- Rendering a new checkout shows the saved payment method

### DIFF
--- a/packages/webcomponents/src/components/checkout/checkout-core.tsx
+++ b/packages/webcomponents/src/components/checkout/checkout-core.tsx
@@ -156,7 +156,7 @@ export class CheckoutCore {
             paymentMethodGroupId={this.checkout?.payment_method_group_id}
             show-saved-payment-methods={!this.disablePaymentMethodGroup}
             bnpl={this.checkout?.bnpl}
-            client-id={this.checkout?.payment_client_id}
+            authToken={this.authToken}
             account-id={this.checkout?.account_id}
             savedPaymentMethods={this.checkout?.payment_methods || []}
             paymentAmount={this.checkout?.payment_amount}
--- a/packages/webcomponents/src/components/checkout/checkout-core.tsx
+++ b/packages/webcomponents/src/components/checkout/checkout-core.tsx
@@ -33,6 +33,7 @@ export class CheckoutCore {
   @State() serverError: string;
   @State() renderState: 'loading' | 'error' | 'success' = 'loading';
   @State() creatingNewPaymentMethod: boolean = false;
+  @State() paymentMethodGroupId: string;
 
   @Event({ eventName: 'submitted' }) submitted: EventEmitter<ICheckoutCompleteResponse>;
   @Event({ eventName: 'error-event' }) errorEvent: EventEmitter<ComponentError>;
@@ -64,6 +65,7 @@ export class CheckoutCore {
       onSuccess: ({ checkout }) => {
         this.checkout = new Checkout(checkout);
         this.renderState = 'success';
+        this.paymentMethodGroupId = this.checkout?.payment_method_group_id;
       },
       onError: ({ error, code, severity }) => {
         this.serverError = error;
@@ -152,6 +154,7 @@ export class CheckoutCore {
             show-card={!this.disableCreditCard}
             show-ach={!this.disableBankAccount}
             show-bnpl={!this.disableBnpl}
+            paymentMethodGroupId={this.paymentMethodGroupId}
             show-saved-payment-methods={!this.disablePaymentMethodGroup}
             bnpl={this.checkout?.bnpl}
             client-id={this.checkout?.payment_client_id}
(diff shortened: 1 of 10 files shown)


packages/webcomponents/src/components/checkout/save-new-payment-method.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  handleCheckboxChange(event: Event) {
+    const target = event.target as HTMLInputElement;
+    this.isChecked = target.checked;
+    this.checkboxChanged.emit(this.isChecked);
+  }
+
+  render() {
+    return (
+      <div class="mt-3 d-flex align-items-center">
```
TiagoFuelber, line 18: I tried with `mt-4` and it looked a little bit more centralized. WDYT?

mt-3: 
image

mt-4: 
image

colganmeanor: Yeah, agreed. It's a little better with `mt-4`. I will update that.

Good catch!

packages/webcomponents/src/components/checkout/new-payment-method.tsx
```diff
@@ -52,7 +62,13 @@ export class NewPaymentMethod {
...
     try {
       const billingFormFieldValues = await this.billingFormRef.getValues();
-      const paymentMethodData = { ...billingFormFieldValues };
+      let paymentMethodData;
+      if (this.saveNewPaymentMethodChecked) {
+        paymentMethodData = { ...billingFormFieldValues, payment_method_group_id: this.paymentMethodGroupId };
+      } else {
+        paymentMethodData = { ...billingFormFieldValues };
+      }
+      console.log('paymentMethodData', paymentMethodData);
```
TiagoFuelber, line 71: Oops

packages/webcomponents/src/components/checkout/new-payment-method.tsx
```diff
@@ -77,6 +93,7 @@ export class NewPaymentMethod {
         </div>
         <h3 class="fs-6 fw-bold lh-lg mb-4">Billing address</h3>
         <justifi-billing-form ref={(el) => (this.billingFormRef = el)} />
+        <justifi-save-new-payment-method hidden={!this.paymentMethodGroupId} />
```
TiagoFuelber, line 96: Nice!! I totally forgot about the `hidden` HTML attribute. 👏

packages/webcomponents/src/components/checkout/checkout-core.tsx
```diff
@@ -155,6 +155,7 @@ export class CheckoutCore {
             show-card={!this.disableCreditCard}
             show-ach={!this.disableBankAccount}
             show-bnpl={!this.disableBnpl}
+            paymentMethodGroupId={this.paymentMethodGroupId}
```
TiagoFuelber, line 158: I think we could add the `payment_method_group_id` to the `Checkout` class constructor then we can pass this prop like: 
`paymentMethodGroupId={this.checkout?.payment_method_group_id}`

Reasons:
* Make it consistent with the backend object entity
* Avoid derived state: line 35 and 69

packages/webcomponents/src/components/checkout/new-payment-method.tsx
colganmeanor: 🙃 🤪

packages/webcomponents/src/components/checkout/checkout-core.tsx
colganmeanor: Ah this is much better. Thankfully it seems we are already passing this in the Checkout constructor - so its simple enough to just remove lines 35 and 69. 😄

TiagoFuelber (approved):
Nice work man! Approved! 💻

TiagoFuelber (dismissed):
Nice work man! New changes make sense to me. And new payment methods are showing up when reload the checkout after saving it. Approved

PR #635, opened 2024-08-14
Ensure insuranceValidation is called before resolvePaymentMethod

### LINKED ISSUE #634: When no insurance selection is made checkout can proceed with BNPL
**Expected Result**
When insurance products are offered in the Checkout component, the behavior should be that it is not accepted or unaccepted for purchase but we ensure that the user picks one or the other before proceeding to pay.

**Actual Result**
Currently, when BNPL/Sezzel is selected as the payment method, no selection is made on the insurance component, clicking "Pay" will open a Sezzel tab and take the user through the flow to pay using BNPL. The insurance component gets validation but at this point the user is already continuing to pay. The other payment methods are working as expected - the "Pay" button will trigger validations and stop payment

colganmeanor:
### Ensure insuranceValidation is called before resolvePaymentMethod

This PR commits the following change:

- Passes `insuranceValidation` into the `resolvePaymentMethod` method so that each of the payment-method components check for an insurance selection. 
- In the case of `new-payment-method.tsx` - the insurance validation is checked along with the payment and billing form validations
- In the case of `sezzle-payment-method.tsx` - it simply checks for an insurance selection before opening the sezzle window
- in the case of `saved-payment-method.tsx` - it simply checks for an insurance selection before posting the checkout completion.

Links
-----

Closes #634 

Developer considerations
--------------

- On any task
  - [x] Perform dev QA

Developer QA steps
--------------------

It would be best to test the checkout component with AND without insurance included just to be safe. 

- [x] New Payment Method - Verify insurance validation fires along payment form / billing form validation. Tokenize call shouldn't happen until all fields are valid. 
- [x] Saved Payment Method - verify insurance validation fires.
- [x] Sezzle Payment Method - verify insurance validation fires before the sezzle window can open

### DIFF
--- a/packages/webcomponents/src/components/checkout/new-payment-method.tsx
+++ b/packages/webcomponents/src/components/checkout/new-payment-method.tsx
@@ -31,7 +31,7 @@ export class NewPaymentMethod {
   }
 
   @Method()
-  async resolvePaymentMethod(insuranceValidation?: any): Promise<PaymentMethodPayload> {
+  async resolvePaymentMethod(insuranceValidation: any): Promise<PaymentMethodPayload> {
     if (!this.paymentMethodFormRef || !this.billingFormRef) return;
 
     const isValid = await this.validate();
--- a/packages/webcomponents/src/components/checkout/new-payment-method.tsx
+++ b/packages/webcomponents/src/components/checkout/new-payment-method.tsx
@@ -31,13 +31,14 @@ export class NewPaymentMethod {
   }
 
   @Method()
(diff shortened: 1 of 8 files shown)


packages/webcomponents/src/components/checkout/new-payment-method.tsx
```diff
@@ -49,6 +50,13 @@ export class NewPaymentMethod {
     }
   }
 
+  async validate(): Promise<boolean> {
```
jakemer, line 53: I love this separation! Nice and clean 🧼

jakemer (approved):
Very nice work. I really like how this turned out!

PR #637, opened 2024-08-15
Improved styling approach and build

jakemer:
- Convert our styles to css and remove the scss build (reduces our build times by a minute or more)
- Create a new `StyledHost` component to be used in place of `Host`. It renders `children` between `Host` tags and injects bootstrap. This should be used on top-level components that have `shadow` set to `true`, and child components should not have shadow DOM so that the styles apply throughout
- Fix radio button default border color on insurance component and checkout payment method options

Links
-----

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
- Radio button border color on season interruption insurance component and the checkout payment method options has been fixed. It was blue by default before and now should be gray or whatever color the theme sets for default.
- Otherwise, components look the same as they did before
   - [ ] With no theme (vanilla bootstrap)
   - [ ] With "Light" theme
   - [ ] With "Dark" theme


TiagoFuelber (approved):
Awesome, so much better. Approved

TiagoFuelber (approved):
✅

PR #641, opened 2024-08-22
Adjustments to SSN and Tax ID fields in PaymentProvisioning component

### LINKED ISSUE #623: Improvements to PaymentProvisioning form
Updates to the following fields in the PaymentProvisioning component:
1. `Tax ID`: 
- update label to `Tax Id (EIN or SSN)`
- remove mask (because tax id  and SSN have different masks)
2. `SSN` field:
- update label to`SSN` 
- keep validation in place: when `ssn_last4` is populated and the SSN field is empty, don't show a validation error.

colganmeanor:
### Adjustments to SSN and Tax ID fields in PaymentProvisioning component

This PR makes the following changes:

- Tax ID Field (Core Info / First form step)
  - label changed from `Tax ID` => `Tax ID (EIN or SSN)`
  - removed mask that was created for Tax ID specificly
  - adds keyDownHandler to prevent alphabet / special character input. Input allows numbers only. 
  - Validation is largely the same, but it no longer checks for the taxIDRegex. 

- SSN Field (Representative and Owner Form Steps)
  - removed logic that changed label for SSN field if ssn_last4 returns a value. Label always shows `SSN`
  - original validation is still in place, so the previous logic still holds true. If ssn_last4 is returned in fetch request, then this field will not be required for this business. 

These are really the only changes worth mentioning - but I did update the files I was working on so that they used single quotes instead of double quotes for strings, etc. 

Links
-----

Closes #623 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA

Developer QA steps
--------------------

The form should pretty much be the same. 

- [x] Label for Tax ID on first form step now says `Tax ID (EIN or SSN)`
- [x] error validation message for Tax ID field should also reflect new label changes - error text says `Enter valid tax id, SSN, or EIN`
- [ ] Tax ID field should largely follow the same validation rules as before. Mask should be removed, but it should still only allow raw number input. Long strings of just a single digit are invalid, as is `123456789`
- [ ] Field does not allow pasting alphabet characters or special characters into the field

- [x] SSN Field on Representative and Owner forms should only show `SSN` as label, no longer shows `Update SSN (optional)` if this form step has already been submitted with an ssn / identification_number
- [x] Validation and logic for this field remains the same. Even though the label is updated, the field should function exactly as befoee

### DIFF
--- a/.changeset/tricky-rocks-relate.md
+++ b/.changeset/tricky-rocks-relate.md
@@ -1,0 +1,7 @@
+---
+"@justifi/webcomponents": minor
+---
+
+- Updated field labels in `PaymentProvisioning` component.
+  - `Tax ID` field label has been updated to `Tax ID (EIN or SSN)`
+  - `SSN` field label no longer changes to `Update SSN (optional)` if submitting this form step again after the first time. 
--- a/packages/webcomponents/src/components/business-forms/owner-form/owner-form.tsx
+++ b/packages/webcomponents/src/components/business-forms/owner-form/owner-form.tsx
@@ -39,10 +39,6 @@ export class BusinessOwnerForm {
     return this.ownerId ? `entities/identity/${this.ownerId}` : 'entities/identity';
   }
 
-  get identificationNumberLabel() {
-    return this.owner.ssn_last4 ? 'Update SSN (optional)' : 'SSN';
-  }
-
   get formTitle() {
     return this.ownerId ? 'Edit Business Owner' : 'Add Business Owner';
   }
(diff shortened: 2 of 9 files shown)


TiagoFuelber:
It's allowing pasting text from the keyboad (CMD + V) 
image

TiagoFuelber:
I was able to type a long string in the SSN field of the General Info form. (But not in the other ones)
image

colganmeanor:
@TiagoFuelber thanks for the comments and feedback! I just pushed up a commit that should fix both of those issues. 

Whenever you have a chance, try again and pasting should only work in that field if your clipboard text has ONLY numbers. If you try to paste alphabetical text it should not allow it. And I also added a max length to this field. 👋🏻

TiagoFuelber (approved):
Nice! Tested again and it's working fine! Approved

Issue #625, opened 2024-08-02
Checkout: Don't show card/bank in payment method group if disabled


jakemer:
The checkout component has `disable-credit-card` and `disable-bank-account` props which will hide the ability to create new credit card and bank account payment methods, however it does not filter those payment methods out of previously saved payment methods. We need to filter the saved payment methods so that we are not showing disabled payment methods types.

Make BNPL the default (after saved payment methods) if it's enabled

jakemer:
This is blocked for the time being - we need the backend to return card/ach types for the saved payment method tokens in order to be able to filter them out when disabled

jakemer:
This issue is dependent on / blocked by https://github.com/justifi-tech/hosted-checkout-be/pull/354

PR #642, opened 2024-08-22
Hide saved disabled payment methods

TiagoFuelber:
**Description**

* This PR makes BNPL the default option after saved payment methods by pushing it before any other payment method in the array. (Jake's PR)
* This also add a new `type` property to the payment method options and filters out the saved payment methods accordingly with `disabled-bank-account` and `disabled-credit-card` props.

Links
-----
#625 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
This can be tested either in the demo-app or in Storybook, just need a checkout with saved credit cards or bank account payment methods
[ ] - passing `disable-credit-card=true` should hide saved credit card payment methods
[ ] - passing `disable-bank-account=true` should hide saved bank account payment methods

### DIFF
--- a/.changeset/purple-dogs-compare.md
+++ b/.changeset/purple-dogs-compare.md
@@ -1,0 +1,5 @@
+---
+"@justifi/webcomponents": patch
+---
+
+Hides saved credit card/bank account payment methods when disable-credit-card or disable-bank-account props are true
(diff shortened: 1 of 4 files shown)


colganmeanor (commented):
Nice work man! 

I just finished testing this in the checkout-demo app. Everything looks good, I did find one minor console error in the dev console when I rendered the checkout with cards and bank accounts disabled. 

Screenshot 2024-08-23 at 11 54 13 AM

Screenshot 2024-08-23 at 11 54 03 AM

Looks like the connectedCallback in the `payment-method-options.tsx` component is erroring out on selecting a payment method if both disable-credit-card and disable-bank-account are `true` 🤔

colganmeanor (approved):

Issue #639, opened 2024-08-15
Handle Checkout status in Checkout component


jakemer:
When the Checkout component fetches the checkout, check the checkout status. If the status is `expired` or `completed` we should emit an error event from the component so the implementer can handle them.

The error events should be distinguishable from each other so that they can be handled differently. For example an implementer might re-route to a receipt if the checkout is `completed` or 404 page if the checkout is `expired`

PR #647, opened 2024-08-27
handle expired and completed checkout statuses

TiagoFuelber:
**DESCRIPTION**
* This PR adds some logic to emit an error event when checkout status is expired or completed

Links
-----
#639 

Developer considerations
-------------

Developer QA steps
--------------------
While we don't have yet `completed` and `expired` statuses on our API yet, it's still possible to test this by overriding the request response on the browser:
* On the network tab, right click the checkout request, and click "Override content"
* Edit the response JSON changing the status to `completed` or `expired` and save.
* An error event should be emitted with error details for each case.

### DIFF
--- a/packages/webcomponents/src/api/Checkout.ts
+++ b/packages/webcomponents/src/api/Checkout.ts
@@ -38,7 +38,8 @@ export interface ICheckout {
   bnpl?: IBnpl;
   total_amount: number;
   insurance_amount: number;
-};
+  status?: string;
+}
 
 export class Checkout implements ICheckout {
   payment_intent_id: string;
@@ -69,6 +70,7 @@ export class Checkout implements ICheckout {
   bnpl?: IBnpl;
   total_amount: number;
   insurance_amount: number;
+  status?: string;
 
   constructor(data: ICheckout) {
     Object.assign(this, data);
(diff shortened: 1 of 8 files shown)


packages/webcomponents/src/api/Checkout.ts
```diff
@@ -38,7 +38,8 @@ export interface ICheckout {
   bnpl?: IBnpl;
   total_amount: number;
   insurance_amount: number;
-};
+  status?: string;
```
jakemer, line 41: Can we do a union type of the actual values from the docs please? https://docs.justifi.tech/api-spec#tag/Checkouts/operation/CreateCheckout

`'created' | 'completed' | 'attempted' | 'expired'`

packages/webcomponents/src/components/checkout/checkout-core.tsx
```diff
@@ -64,6 +64,22 @@ export class CheckoutCore {
     this.getCheckout({
       onSuccess: ({ checkout }) => {
         this.checkout = new Checkout(checkout);
+        const { status } = this.checkout;
+        if (status === 'completed' || status === 'expired') {
```
jakemer, line 68: Maybe we can just emit all statuses since the integrator will need to check the status to decide how to handle it anyways. It would remove some complexity on our side. What do you think?

TiagoFuelber: That's a good ideia!

jakemer (approved):

Issue #645, opened 2024-08-26
PaymentsList and PayoutsList: Ability to select date ranges


jakemer:
Add date range filtering to the `PaymentsList` and `PayoutsList`

PR #653, opened 2024-08-28
add date range filter on payments/payouts table

TiagoFuelber:
* This adds two date inputs for selecting a date range for filtering payments-list and payouts-list

**NOTE**: The selected dates are excluded from the results. For example, if the start_date is set to the 24th, the results will exclude data from the 24th and only include data from the 25th onward. I believe this is the current behavior on the backend. Should this be changed?

image

Links
-----
#645 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [X] Does the component have exportedparts for styling? If so, add unit a unit test for each exportedpart

Developer QA steps
--------------------
[ ] - Payments List and Payouts List components should now display two datepickers for filtering by a date range


packages/webcomponents/src/components/payments-list/payments-list-core.tsx
```diff
@@ -71,9 +71,29 @@ export class PaymentsListCore {
...
+  }
+
   render() {
     return (
       <Host>
+        <div class="row gy-3 mb-4">
+          <div class="col-2">
+            <form-control-date
+              name="created_after"
+              label="Date from:"
```
jakemer, line 85: I am not sure yet, but I think we should label them `Start Date` and `End Date`, with no `:`

jakemer (dismissed):
Nice work!

jakemer (approved):

Issue #654, opened 2024-08-29
Ability to export payout detail data


TiagoFuelber:
- Add an export to `PayoutDetail` component
- Add an export to `PayoutsList` component as a row action

jakemer:
https://docs.justifi.tech/api-spec#tag/Payouts/operation/UpdatePayout

PR #656, opened 2024-09-04
add download CSV button for payout details and list

TiagoFuelber:
**Description**

* Added a "Download CSV" button to the payout details page, allowing users to download the payout details as a CSV file.
* Introduced a new "Actions" column in the payouts list table, which includes the "Download CSV" button for each entry.

Links
-----
#654 

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------

[ ] - PayoutsList component: It should have a new "Actions" column, and a download icon. Clicking it, should download a CSV file with Payout details.
[ ] - PayoutDetails component: It should have a "Download CSV" button beside the created, updated and ID infos. This should download a CSV as well.


packages/webcomponents/src/components/payout-details/payout-details-core.tsx
```diff
@@ -7,10 +7,12 @@ import StyledHost from '../../utils/styled-host/styled-host';
 
 @Component({
   tag: 'payout-details-core',
+  styleUrls: ['../../styles/icons.css'],
```
TiagoFuelber, line 10: @jakemer should I move this file to the root.css file?

jakemer: I don't think it'll work to move these styles to root.css because it needs a way (such as a css part) to make it through to the shadow DOM. I am wondering, though, if it would make sense to have an actual SVG file with the dimensions set on it and import it directly into the component rather than using css/background-image? Or is there a bootstrap native way to do icons?

jakemer (approved):

PR #667, opened 2024-09-06
Payment provisioning custom style cleanup

jakemer:
The idea with this PR is to begin to organize shared "ui components" under a single directory, and to help us move towards our goal of removing CSS variables and using CSS parts only. Some of our remaining custom styling is for submit and loading buttons, which simply applies css variables to certain bootstrap buttons. By making a Button ui component we can apply css variables to that instead and move it to a central location where it can easily be deleted when the time comes to remove css variables entirely.

- Make a Button UI component which abstracts bootstrap classes and `exportparts`
- Use the button in `PaymentProvisioning`, `PaymentForm`, and `Checkout`
- Remove styles from `PaymentProvisioning`, `PaymentForm` and `Checkout` that apply css variables to buttons, and instead centralize the styles in `root.css` using css part selectors 
- Create a `ui-components` directory for the new `Button` component and also move `ButtonSpinner` and `StyledHost` to it

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
Chromatic should mostly cover these changes as we need to make sure there are not visual regressions, but it would still be good to check the following in storybook and make sure the styling looks correct (in the themes too if the component has them)
- PaymentProvisioning
- PaymentForm
- Checkout


apps/docs/stories/components/PaymentForm/index.stories.tsx
```diff
@@ -118,23 +123,16 @@ const meta: Meta = {
...
 
 const Template = (args: any) => {
   // The <div> here should be replaced by a `display` property in the cardForm potentially
   return `
-    <div>
-      <justifi-payment-form
-        ${getAttributesString(args)}
-      />
-    </div>
+    <justifi-payment-form ${getAttributesString(args)}></justifi-payment-form>
```
colganmeanor, line 133: @jakemer not sure where the issue is in this file - but I'm seeing the new Theme selector that works exactly as the Checkout story, and I'm also seeing an additional select control called `theme` (lowercase) that seems to be an accidental addition. This additional control doesn't have the same options, and doesn't actually change anything. But we should fix that. 

But otherwise the dark and light themes look good on the PaymentForm. 👍🏻

packages/webcomponents/src/components/business-forms/owner-form/owner-form.tsx
```diff
@@ -299,7 +299,7 @@ export class BusinessOwnerForm {
                   class={`btn btn-primary jfi-submit-button${this.isLoading ? ' jfi-submit-button-loading' : ''}`}
                   onClick={() => this.handleAddOwner()}
                   disabled={this.isLoading}>
-                  {this.isLoading ? LoadingSpinner() : this.submitButtonText}
+                  {this.isLoading ? ButtonSpinner() : this.submitButtonText}
```
colganmeanor, line 302: I see you added in the ButtonSpinner here, but do you want to use the new Button component you made as well? Then you dont't need to import the ButtonSpinner and can just pass this.isLoading to it?

packages/webcomponents/src/ui-components/styled-host/styled-host.tsx
```diff
@@ -2,13 +2,11 @@ import { FunctionalComponent, HostAttributes } from "@stencil/core/internal";
 import { Host, h } from '@stencil/core';
 import modifiedBootstrap from './modified-bootstrap.css';
 
-const StyledHost: FunctionalComponent<HostAttributes> = (props, children) => {
+export const StyledHost: FunctionalComponent<HostAttributes> = (props, children) => {
   return (
-    <Host {...props} class="body">
+    <Host {...props} class="body" style={{ display: 'block' }}>
```
colganmeanor, line 7: this file is doing so much heavy lifting for us now. Great job with these abstractions!

colganmeanor (dismissed):
Left a couple of tiny comments, but overall great work @jakemer !

I feel like this PR sets us up very nicely for a handful of great revisions / refactors - definitely nice to see on a Friday! 

Approved and dev QA approved 👍🏻

packages/webcomponents/src/ui-components/button.tsx
```diff
@@ -0,0 +1,44 @@
...
+    secondary: ['button', 'button-secondary'],
+  };
+
+  const getCSSParts = () => {
+    const baseCSSParts = exportPartMapping[props.variant];
+    if (props.isLoading) baseCSSParts.push('button-loading');
+    return baseCSSParts.join(' ');
+  };
+
+  console.log(props.clickHandler);
```
colganmeanor, line 30: Oh and also spotted a rogue console.log to remove 👀

packages/webcomponents/src/components/business-forms/owner-form/owner-form.tsx
jakemer: Yes, we should do that eventually but it felt like a slippery slope - I think we can have a PR that just replaces all buttons with the new one. We will have to expand on the Button component to handle other styles of buttons such as `btn-danger`

colganmeanor (approved):

packages/webcomponents/src/ui-components/button.tsx
```diff
@@ -3,14 +3,14 @@ import { ButtonSpinner } from './button-spinner';
...
-  type: 'button' | 'submit' | 'reset';
-  clickHandler: (e: any) => void;
-  isDisabled?: boolean;
   isLoading?: boolean;
-  isHidden?: boolean;
 }
 
-export const Button: FunctionalComponent<ButtonProps> = (props, children) => {
+type ButtonElementProps = ButtonProps & {
+  [key: string]: any; // This allows any additional properties
```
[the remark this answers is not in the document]
jakemer: Exactly! 😄

colganmeanor (approved):

PR #672, opened 2024-09-09
Remove PaymentForm, CardForm, and BankAccountForm from Storybook

jakemer:
We have several components in Storybook currently which are used in payments processing, however our recommended route is to use the Checkout component going forward. Therefore we would like to not have other payment/payment method related components in Storybook as it causes confusion and documentation overhead.

Developer considerations
--------------

- On any task
  - [ ] Perform dev QA
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------


TiagoFuelber:
@jakemer Just a question I have, will the PaymentForm be still available, will it be deleted?

TiagoFuelber (approved):
Approved

Issue #689, opened 2024-09-18
SPIKE: web-component-library environments


TiagoFuelber:
- How can we make the components use the staging API environment?
- How can we make them use staging iframe-components (we first need to deploy a staging version of that app)

PR #696, opened 2024-09-24
689 spike web component library environments

TiagoFuelber:
**PR Overview**
This PR introduces environment variables to the build process, ensuring that:
- Staging APIs are used when running locally 
- Mocks are used in Storybook production.
- Mocks are used in Chromatic
- Prod API is used on the publish NPM package

**Local Setup**
To run the project locally, you’ll need to create the following file:

`packages/webcomponents/.env` with the content:

```
IFRAME_ORIGIN=https://components.justifi-staging.com/v1
PROXY_API_ORIGIN=https://wc-proxy.justifi-staging.com
```

These variables have also been added to GitHub Secrets, making them available during deployment.

**Key Changes**
* dotenv integration: Environment variables are now loaded from an .env file using the dotenv library.
* Stencil build workaround: Since `process.env` is not accessible in the Stencil build that runs on a browser environment, the `rollup/replace` plugin is used to replace all occurrences of the variables on compile time with their correct values during the build.
* Test setup: A `setupTests.j`s file was added to set the variables in test environments, as `process.env` is not directly available in test files.
* E2E tests: Removed the e2e test files as they were just 3 and were testing very basic stuff.

**Important Note**
This update switches the project to use the backend staging environment when running locally. As a result, new IDs and auth tokens will need to be configured in the environment to successfully load data for all components.

Links
-----
#689 

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
[ ] - When running Storybook locally, all components should call `https://wc-proxy.justifi-staging.com` and the iframe-origin on payment-method components should be `https://components.justifi-staging.com/v1`


.github/workflows/chromatic-deployment.yaml
```diff
@@ -27,6 +27,20 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
+      - name: Create .env file from secrets
+        shell: bash
+        working-directory: ./apps/docs
+        run: |
```
jakemer, line 33: Can some of this be abstracted and used across github actions? If not now, in the future?

packages/webcomponents/stencil.config.ts
```diff
@@ -18,6 +22,19 @@ export const config: Config = {
...
+    }),
+  ],
   validatePrimaryPackageOutputTarget: true,
   sourceMap: false,
+  testing: {
+    setupFiles: ['./setupTests.js'],
+  },
+  buildEs5: false,
+  // Setting the minify flags to false will help retain console.logs
+  minifyJs: false,
```
jakemer, line 38: Will the production build not have minified js and css then? That will make load times much longer

TiagoFuelber: Good catch, I changed this to false just for debugging reasons and forgot to roll it back.

.github/workflows/chromatic-deployment.yaml
TiagoFuelber: Yeah, I think so.
I'll add this suggestion to the CI refactor task that we already have in the backlog 👍

jakemer (approved):

Issue #662, opened 2024-09-05
Refactor business payload parsing file


TiagoFuelber:
Right now, the Business Payload we send to back end is just the same we receive from API, but parsing out the properties the BE doesn't accept. `packages/webcomponents/src/components/business-forms/utils/payload-parsers.ts`

We should instead, create the payload object with the properties we need to send. This way, if BE adds a new property, the request will not break.

PR #697, opened 2024-09-26
Remove payload parsers and add payload getter to classes

TiagoFuelber:
**Summary**
This PR refactors the payload handling logic by replacing the payload-parsers file with a payload getter method in the relevant business classes.

**Problem**
The previous payload-parsers approach removed properties not accepted by our API. This method was error-prone and not maintainable, as it required constant updates whenever new properties were introduced by the back-end, potentially causing bugs.

**Solution**
The new solution introduces a payload getter method in the business-related classes. This method explicitly returns only the properties that are accepted by our API, ensuring that payloads are always valid.

**Affected components**
- Business Form
- Payment Provisioning Form

Links
-----
#662 

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
[ ]  - Business Form and Payment Provisioning should work as before.


packages/webcomponents/src/components/business-forms/utils/payload-parsers.ts
```diff
@@ -1,48 +0,0 @@
-// These props should not be sent to the server
```
jakemer, line 1: Awesome to be able to delete this!

packages/webcomponents/src/components/business-forms/payment-provisioning/document-uploads/business-document-upload-form-step.tsx
```diff
@@ -59,7 +59,7 @@ export class BusinessDocumentFormStep {
     this.formLoading.emit(true);
     try {
       const response: IApiResponse<IBusiness> = await this.api.get(this.businessEndpoint);
-      this.business = { ...new Business(response.data) };
+      this.business = { ...new Business(response.data) } as Business;
```
jakemer, line 62: I'm sure there is a reason, but why can't this be `new Business(response.data)` without spreading and casting? Does it need to be spread onto a new object to trigger a re-render? If so it's kinda odd, I'd think that the constructor for `Business` would return a new object and trigger a re-render too 🤔

jakemer (dismissed):

packages/webcomponents/src/components/business-forms/payment-provisioning/document-uploads/business-document-upload-form-step.tsx
TiagoFuelber: Yes, I agree it’s a bit odd. The original code was written this way because simply using new Business was causing a bug at the time, and this was the solution that worked. I’ll test using new Business again to see if the issue persists. If it does, I’ll add a comment explaining why this approach is necessary

colganmeanor: Oh hey I have some context to add here: 

The reason we instantiated and spread like this is because when we would send the instantiated data without spreading it, we would fail the POST request because for some reason the constructor of the class was being sent along with the payload.... It was a weird bug. At the time we just decided we were fine with spreading the instantiated class and just moving on. 

Maybe we can fix this properly now that we are sending the results of the class getter method as the payload, instead of the entire class / object itself. 🤔

colganmeanor (changes requested):

packages/webcomponents/src/components/business-forms/payment-provisioning/document-uploads/business-document-upload-form-step.tsx
TiagoFuelber: Oh exactly, thanks Colgan. That was the case. Now with the new approach, I tested it and the spread is not necessary anymore! 🙌 💯

packages/webcomponents/src/components/business-forms/business-form/business-form.tsx
```diff
@@ -80,7 +79,7 @@ export class BusinessForm {
     try {
       const values = this.formController.values.getValue();
       const initialValues = this.formController.getInitialValues();
-      const payload = parseBusiness(values, initialValues);
+      const payload = new Business({ ...initialValues, ...values }).payload;
```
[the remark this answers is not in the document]
TiagoFuelber: I am not sure if I got what you mean! The payload of the Business class is returning everything and it's being used on the original Business form. For the Payment Provisioning first step, there is the `CoreBusinessInfo` which returns a smaller payload.

packages/webcomponents/src/api/Business.ts
```diff
@@ -24,6 +24,7 @@ export interface IAddress {
   country?: string;
   created_at?: string;
   updated_at?: string;
+  payload?: any;
```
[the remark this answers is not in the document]
TiagoFuelber: Good catch! I added this because at some point TS was complaining, but in the end this was not needed anymore. 👍

packages/webcomponents/src/components/business-forms/business-form/business-form.tsx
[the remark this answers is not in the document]
colganmeanor: You're right! My bad, I didn't realize when I was looking earlier that it was already covered. Good catch!

packages/webcomponents/src/api/Identity.ts
```diff
@@ -64,6 +68,23 @@ export class Owner implements Identity {
     this.title = owner.title;
     this.updated_at = owner.updated_at;
   }
+
+  // Getter method to return the payload in the desired shape
+  public get payload() {
```
colganmeanor, line 73: One small thing - we need to add `identification_number` to the payload getter for both Representative and Owner.  This is the property that posts the SSN for the data. 👍🏻

colganmeanor (approved):

PR #698, opened 2024-09-27
Update Api.ts to stringify sent payloads

jakemer:
Update our `Api.ts` so that the methods which send data stringify it, rather than doing so in each implementation

Links
-----

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------


colganmeanor (dismissed):
Everything looks good, nice ccleanup!

I was testing and noticed that the individual owner-form component needed to get the same changes, since I was testing and it failed to post or patch owner data for an individual form. 

Just pushed up a commit that fixed that, and otherwise I say its good to go. ✅ 🙌🏻

colganmeanor (approved):

Issue #500, opened 2024-05-09
Notify users about web-components new versions


TiagoFuelber:
On component load, check what the latest version on NPM is and compare to the installed version. If the component is old (maybe specifically for major or minor version differences?) log to the console that there is a newer version.

PR #699, opened 2024-09-30
Let users know when there's a new major version

TiagoFuelber:
Introduce a function to detect new major versions and log a console warning to notify users.

Links
-----
#500 

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
[ ] - Change the package.json version to be minor than 4, then a warning message should be logged in the console.


colganmeanor (approved):
I think technically everything is great! 

I just have a question though - when we are console logging the newest version from NPM, do we want to include `rc` releases/versions? 

Or should we try to strip out this part of the text and only show the non-rc releases?

@TiagoFuelber @jakemer what do you think? 

Screenshot 2024-09-30 at 3 40 21 PM

jakemer (approved):

PR #709, opened 2024-10-16
Refactored the Checkout to be styled with CSS parts

TiagoFuelber:
This pull request adds the necessary `parts` so the Checkout can be styled without the use of CSS variables. (Some variables are still needed for the payment method on the iframe)

Links
-----
#707 

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------

As a suggestion for QA, I would recommend to open Storybook prod side by side with this branch locally.

[ ] - Checkout appearance should be the same as before in all themes (default Bootstrap, Light and Dark)
[ ] - All different states should also have not changed. (Input focus, radio buttons checked, invalid inputs etc.)


apps/docs/stories/themes/dark.ts
```diff
@@ -37,6 +37,66 @@ justifi-payment-provisioning {
...
+::part(radio-input-checked):focus {
+  border-color: #ff785a;
+  box-shadow: 0 0 0 0.25rem rgba(13, 110, 253, 0.4);
+}
+
+::part(radio-input-label) {
+  color: #fff;
+}
+
+::part(payment-method-header):hover {
```
jakemer, line 72: Can we get rid of this in a later PR and use something more standard?

apps/docs/stories/themes/dark.ts
```diff
@@ -93,66 +150,7 @@ justifi-payment-provisioning {
   --jfi-error-message-margin: 5px;
   --jfi-error-message-font-size: 12px;
 
-  /* Below only used in justifi-payment-form */
```
jakemer, line 96: Love to see all this red!

packages/webcomponents/src/components/checkout/checkout-core.tsx
```diff
@@ -194,15 +193,15 @@ export class CheckoutCore {
         <div class="row gy-3 jfi-checkout-core">
           <div class="col-12">
             {/* componentize this */}
-            <h2 class="fs-5 fw-bold pb-3 jfi-header">Summary</h2>
+            <h2 class="fs-5 fw-bold pb-3 jfi-header" part="header-2">Summary</h2>
```
jakemer, line 196: Can we remove the `jfi-header` and other custom classes in this template?

packages/webcomponents/src/components/checkout/checkout.tsx
```diff
@@ -9,6 +9,7 @@ import { checkPkgVersion } from '../../utils/check-pkg-version';
 
 @Component({
   tag: 'justifi-checkout',
+  shadow: true,
```
jakemer, line 12: I think we should make it so that `StyledHost` has shadowDom so we dont have to worry about being consistent about where we put this

packages/webcomponents/src/components/checkout/new-payment-method.tsx
```diff
@@ -108,7 +108,9 @@ export class NewPaymentMethod {
       <div class="payment-method">
         <div
           class={`payment-method-header p-3`}
```
jakemer, line 110: Can we get rid of the custom class here now?

packages/webcomponents/src/components/checkout/new-payment-method.tsx
```diff
@@ -108,7 +108,9 @@ export class NewPaymentMethod {
...
-          onClick={() => this.onPaymentMethodOptionClick()}>
+          onClick={() => this.onPaymentMethodOptionClick()}
+          part="payment-method-header"
```
jakemer, line 112: I think we should give this a more generic name in case we do this ui pattern somewhere else. Such as `radio-list` and `radio-list-header`, or maybe there is a more standard bootstrap way of accomplishing this

PR #757, opened 2024-11-19
New Component: TokenizePaymentMethod

jakemer:
The `TokenizePaymentMethod` web component will be used for our customers with subscription style use cases as we will be deprecating `PaymentForm` in version `5.0.0`. `TokenizePaymentMethod` almost entirely re-uses code from `Checkout`, has a more appropriate name, and styling can be applied to it in a way that is consistent with `Checkout`.

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------

Pull the example from https://github.com/justifi-tech/web-component-library/pull/757

- Run `npm start:tokenize-payment-method`
- Click the "submit" button without entering any info. It should show inline errors.
- Fill out the fields using valid test card data and click "submit" again. The output pane in the example should show the created token

External submit button
- Modify the code in the example to:
  - Hide the submit button `hide-submit-button="true"`
  - Create an external button with an id
  - Create a click handler for the external button that calls `tokenizePaymentMethod`
  - click the button to tokenize, and the token should be output to the pane just as it did in the original example


colganmeanor (dismissed):
Looks very clean! 

Great work! 🦖

colganmeanor:
I'm finished with my QA. Everything looks great! 

I tested all events, props, and the fillBillingForm method and all work as expected and documented. Good work! 

I think one thing we're missing here is the addition of some kind of mock data response json file, and a few lines additions to the `mockAllServices.ts` file that handles the Mirage passthrough logic. That would be helpful for us on Storybook so folks aren't testing real API calls with this component example in storybook. 

And maybe some unit tests?

colganmeanor (approved):

PR #761, opened 2024-11-26
Refactor Web Component Styling Approach

TiagoFuelber:

- Delete the `root.css` file and adds some logic to handle the iframe input styles (focused and invalid) using a hidden input component. 
- Update the docs removing references and instructions about CSS variables
- Delete CSS files for individual components
- Adds the `billing-form` part, so it's possible to hide it using CSS.

Links
-----

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------


packages/webcomponents/src/ui-components/form/hidden-input.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return new Promise((resolve, _reject) => {
+      this.hiddenInput.focus();
+      setTimeout(() => {
+        this.hiddenInput.blur();
+        let computedStyles = getComputedStyle(this.hiddenInput);
+        resolve({
+          boxShadow: computedStyles.boxShadow,
+          border: computedStyles.border,
+        });
+      }, 1000);
```
jakemer, line 28: Can we see if we can reduce the time here?

packages/webcomponents/src/ui-components/form/hidden-input.tsx
```diff
@@ -0,0 +1,63 @@
...
+        let computedStyles = getComputedStyle(this.hiddenInput);
+        this.hiddenInput.blur();
+        this.hiddenInput.classList.remove('is-invalid');
+        const styles = {
+          boxShadow: computedStyles.boxShadow,
+          border: computedStyles.border,
+        };
+
+        resolve(styles);
+      }, 1000);
```
jakemer, line 46: And here? If this happens to be related to bootstrap animations, maybe we can remove the animation on the hidden input to get this as close to a 0 timeout as possible

packages/webcomponents/src/ui-components/form/hidden-input.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  render() {
+    return (
+      <Host>
+        <input
+          ref={el => this.hiddenInput = el}
+          type="text"
+          class="form-control"
+          tabindex="-1"
+          style={{ height: '0', opacity: '0', pointerEvents: 'none' }}
```
jakemer, line 58: Maybe animations can be disabled here

packages/webcomponents/src/ui-components/form/iframe-input.tsx
```diff
@@ -10,23 +11,18 @@ import { MessageEventType } from "../../components/checkout/message-event-types"
...
   private frameService: FrameCommunicationService;
-  private hiddenInput!: HTMLInputElement;
 
   @Element() el: HTMLElement;
 
   @State() isFocused: boolean = false;
   @State() isValid: boolean = true;
   @State() errorText: string;
-  @State() fontFamily: string;
+  @State() fontFamily: string = 'Arial';
```
jakemer, line 20: Not sure but maybe `'Sans serif'` would be a better fallback

packages/webcomponents/src/ui-components/form/iframe-input.tsx
```diff
@@ -83,19 +79,32 @@ export class IframeInput {
...
       <Host class="form-group d-flex flex-column">
-        {/* hidden input with bootstrap styles so we can get the computed fontFamily and pass it to the iframe */}
-        <input ref={el => this.hiddenInput = el} type="text" class="form-control" style={{ display: 'none' }} />
-        <label htmlFor="" part="label">{this.label || ''}</label>
-        <div part={this.part}>
+        <label
+          class="form-label"
+          htmlFor=""
+          part="label"
+          style={{ height: '20px', minHeight: '20px' }}
```
jakemer, line 99: Just wondering why the `height` and `minHeight` need to be overridden?

packages/webcomponents/src/ui-components/form/iframe-input.tsx
```diff
@@ -105,7 +114,7 @@ export class IframeInput {
                 this.iframeElement = el as HTMLIFrameElement;
                 this.initializeFrameCommunicationService();
               }}
-              height="100%"
+              height="20px"
```
jakemer, line 117: Why `20px`? Just curious

packages/webcomponents/src/utils/utils.ts
```diff
@@ -83,16 +89,49 @@ export function formatAddress(address: Address): string {
...
-  const computedStyles = getComputedStyle(document.body);
-  return (
-    computedStyles
-      ?.getPropertyValue('--jfi-load-google-font')
-      ?.trim()
-      .replace(/'|"/g, '')
-      .replace(' ', '+') || null
-  );
-}
+export const MapPaymentStatusToBadge = (status: string) => {
```
jakemer, line 92: I think this should go somewhere else, like in a payments specific file. That may have even been done already in another branch... may want to check with @colganmeanor

packages/webcomponents/src/utils/utils.ts
```diff
@@ -150,34 +189,6 @@ export function composeQueryParams(values: string[]) {
   return queryParams.join('');
 }
 
-export async function loadFontsOnParent() {
```
jakemer, line 153: Awesome, glad we can delete this now 👍

jakemer (commented):
Very excited about this PR - it cleans a lot of things up! Nice1

packages/webcomponents/src/utils/utils.ts
TiagoFuelber: This mapping function was already here before this PR.
I think if this was not moved to another file yet we can do it in another PR

colganmeanor: @TiagoFuelber @jakemer good call - 

I did update this in `main` actually when we merged in the Terminals List component into main. And I believe that update has since been merged back into `release-5.x.x` so this shouldn't be an issue anymore as long as this branch is up to date with `release-5.x.x` ? 🤔

PR #768, opened 2024-12-02
Update bank account form schema

### LINKED ISSUE #767: FIX: Loosen validation requirements for account owner in `payment-provisioning / bank-account-step`
Request from client to loosen validation requirements in `payment-provisioning` component's bank account step

We currently have regex validations in place that restrict the `account_owner` field to be a string with only alphabetical characters. Clients may need to use this field to list the name of a business instead of a person. 

Let's ensure we are using the same regex validation we use to validate business names for this field.

colganmeanor:
### Update bank account form schema

This PR updates the schema for the bank account step in `payment-provisioning`. 

- `account_owner` field is no longer limited to just alphabetical characters. Users can now provide numbers and special characters: `,`, `.`, `&`, `-`

Links
-----

Closes #767 

Developer QA steps
--------------------

To QA this PR you will need to render the `payment-provisioning` component and proceed to the bank account form step (step 6). We are testing the `account_owner` field only.

- [x] Verify that field is no longer limited to letters only. This field should successfully accept strings with numbers, or the special characters: `,`, `.`, `&`, `-`
  - [x] example string to test: `Accessible Furniture & Decor - Atlanta-Sandy Springs, Inc. 123456`

### DIFF
--- a/.changeset/flat-students-develop.md
+++ b/.changeset/flat-students-develop.md
@@ -1,0 +1,6 @@
+---
+"@justifi/webcomponents": patch
+---
+
+- Updated validation for account owner input field in bank account form step of `justifi-payment-provisioning` component. 
+  - In addition to alphabetical characters, this input will now accept numbers and certain special characters: `,`, `.`, `&`, and `-`
--- a/.changeset/flat-students-develop.md
+++ b/.changeset/flat-students-develop.md
@@ -3,4 +3,4 @@
 ---
 
 - Updated validation for account owner input field in bank account form step of `justifi-payment-provisioning` component. 
-  - In addition to alphabetical characters, this input will now accept numbers and certain special characters: `,`, `.`, `&`, and `-`
+  - In addition to alphabetical characters, this input will now accept numbers and certain special characters: `,`, `.`, `'`, `:`, `&`, and `-`
--- a/.changeset/flat-students-develop.md
+++ b/.changeset/flat-students-develop.md
@@ -2,5 +2,4 @@
 "@justifi/webcomponents": patch
 ---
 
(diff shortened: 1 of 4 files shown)


Claudia108:
Oh yes, good idea!

Claudia108 (dismissed):
Looks good to me!

TiagoFuelber:
I asked ChatGPT what would be the typically disallowed characters for this case and it answered this: 

image

I tried entering some special characters, and the field accepted them. Do you think we should consider updating the regex to explicitly disallow those characters?

TiagoFuelber:
> > I asked ChatGPT what would be the typically disallowed characters for this case and it answered this:
> > image
> > I tried entering some special characters, and the field accepted them. Do you think we should consider updating the regex to explicitly disallow those characters?
> 
> @TiagoFuelber good thinking to ask chat gpt! I tried testing out the current changes in this PR with all the special characters you listed here - none of them were actually accepted, except for the ones that we have cased/allowed in `businessNameRegex` from `schema-helpers.ts`.
> 
> Are you sure you were able to pass the validation with any of these special characters? I cannot recreate this on my end. Are you sure you were testing with the `Account Owner Name` field in the bank_account form step? 🤔
> 
> That said - I am totally cool with omitting `'`, and `"` from allowed characters. I don't think that would mess anything up since we are stringifying the payload before sending to the BE, however I would rather err on the side of caution here...
> 
> @Claudia108 Based on Tiago's comment I am going to remove `'` from our regex for this field.

Hey,  yes, I'm running the branch locally and could type those characters here

image

That's weird.

jakemer (approved):

PR #778, opened 2024-12-10
Improve date filter behavior

colganmeanor:
Links
-----

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------

### DIFF
--- a/packages/webcomponents/src/components/payments-list/payments-list-filters.tsx
+++ b/packages/webcomponents/src/components/payments-list/payments-list-filters.tsx
@@ -1,6 +1,7 @@
 import { Component, h, Prop } from '@stencil/core';
 import { debounce } from 'lodash';
 import { PaymentsParams } from '../../api';
+import { convertToUTC, convertUTCToLocal } from '../../utils/utils';
 
 @Component({
   tag: 'payments-list-filters'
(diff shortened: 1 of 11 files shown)


packages/webcomponents/src/components/payments-list/payments-list-filters.tsx
```diff
@@ -10,13 +11,21 @@ export class PaymentsListFilters {
...
+  @State() createdBeforeInputValue: string;
+
   private debouncedSetParamsOnChange: (name: string, value: string) => void;
 
   componentWillLoad() {
     // debounced input handler for text input
     this.debouncedSetParamsOnChange = debounce(this.setParamsOnChange, 300);
   }
 
+  handleDateInput = (name: string, value: string) => {
```
TiagoFuelber, line 24: It looks like it's working in both payments and payouts list, but in `payouts-list-filters` this function is setting the `createdAfterInputValue` and `createdBeforeInputValue` state. 
Should we also set this in here?

TiagoFuelber (approved):
Looks much better now! 💯

packages/webcomponents/src/components/payments-list/payments-list-filters.tsx
colganmeanor: Thanks @TiagoFuelber - after this comment I realized we don't actually need these state variables at all in this file. So its a bit simpler now

jakemer (approved):

PR #795, opened 2024-12-19
CSS parts overhaul

TiagoFuelber:
**Summary of the changes**

- Improvements on radio/checkbox inputs and skeleton styles in dark and light themes
- Added a skeleton to the business details (skeleton for other components will be added in a different PR)
- Refactored the skeleton component removing the variants prop, for the reason that it was not very helpful.
- Fixed some form-control-radio weird behaviors 
- Added some new parts (textPrimary, textSecondary, heading4 and heading5)

Links
-----

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------

- [x] [Payment Provisioning Form] - All radio button should look and work nice in all three themes (Bootstrap, Light, Dark)
- [X] The tooltips should look nice in all three themes.
- [X] Business Details should display a Skeleton when loading.


packages/webcomponents/src/ui-components/form/form-control-radio.tsx
```diff
@@ -61,46 +57,48 @@ export class RadioInput {
     if (this.errorText) {
       return inputRadioInvalid;
     }
-    if (this.isFocused) {
-      if (this.checked) {
-        return inputRadioCheckedFocused;
-      }
-      return inputRadioFocused;
+
+    if (this.isFocused && this.checked) {
```
jakemer, line 61: I like this better than the nested condition 👍

packages/webcomponents/src/ui-components/form/form-control-radio.tsx
```diff
@@ -61,46 +57,48 @@ export class RadioInput {
...
-            </label>
-          </div>
-          <form-control-help-text helpText={this.helpText} name={this.name} />
-          <form-control-error-text errorText={this.errorText} name={this.name} />
+      <Host class='form-group d-flex flex-column'>
+        <div class="form-check">
+          <input
+            ref={el => (this.radioElement = el as HTMLInputElement)}
+            type="radio"
+            id={`${this.name}+${this.value}`}
```
jakemer, line 83: Nice, makes sense

packages/webcomponents/src/ui-components/skeleton.tsx
```diff
@@ -5,59 +5,33 @@ interface SkeletonProps {
   component?: string;
   height?: string | number;
   width?: string | number;
-  variant?: 'circular' | 'rectangular' | 'rounded' | 'text';
```
jakemer, line 8: Is there no need for variants anymore?

jakemer (approved):

packages/webcomponents/src/components/business-details/business-details-loading.tsx
```diff
@@ -0,0 +1,33 @@
...
+          <Skeleton />
+          <Skeleton />
+        </div>
+      </div>
+    </div>
+  );
+
+  return (
+    <Fragment>
+      <Block />
```
jakemer, line 27: Nice 👍

jakemer (approved):

jakemer (approved):

colganmeanor:
I'm seeing some quirky behavior with the checkbox in the Light theme - but it works great in dark theme and bootstrap default! 


https://github.com/user-attachments/assets/18295776-fd96-4d11-b5e5-387cb35120dd

packages/webcomponents/src/components/business-details/business-details-core.tsx
```diff
@@ -52,7 +52,11 @@ export class BusinessDetailsCore {
 
   render() {
     if (this.renderState === RENDER_STATES.LOADING) {
-      return <StyledHost><Spinner /></StyledHost>;
+      return (
+        <StyledHost>
+          <BusinessDetailsLoading />
```
colganmeanor, line 57: This looks so much nicer!!

packages/webcomponents/src/components/checkout/new-payment-method.tsx
```diff
@@ -86,7 +86,8 @@ export class NewPaymentMethod {
     }
   }
 
-  onPaymentMethodOptionClick = () => {
+  onPaymentMethodOptionClick = (e) => {
+    e.preventDefault();
```
colganmeanor, line 90: just curious about the reason for this change?

packages/webcomponents/src/components/checkout/saved-payment-method.tsx
```diff
@@ -22,30 +22,27 @@ export class SavedPaymentMethod {
...
-            <form-control-radio
-              name="paymentMethodType"
-              id={this.paymentMethodOption?.id}
-              value={this.paymentMethodOption?.id}
-              checked={this.isSelected}
-              inputHandler={() => this.onPaymentMethodOptionClick()}
-              label={`${CardBrandLabels[this.paymentMethodOption?.brand]} *${this.paymentMethodOption?.acct_last_four}`}
-            />
-          </div>
+      <Host class="payment-method">
```
colganmeanor, line 32: Just curious about why we are using the `Host` element from Stencil here and not our `StyledHost`?

packages/webcomponents/src/ui-components/form/form-control-radio.tsx
```diff
@@ -29,78 +28,73 @@ export class RadioInput {
...
-    this.updateInput(this.value);
-  }
-
   handleFormControlInput = (event: any) => {
     const target = event.target;
     const name = target.getAttribute('name');
-    this.inputHandler(name, target.value);
-    this.formControlInput.emit({ name, value: target.value });
+    this.inputHandler(name, this.value);
+    this.formControlInput.emit({ name, value: this.value });
```
colganmeanor, line 43: This is unrelated to this specific PR - but I wanted to note it because `handleFormControlInput` can probably be shortened to something like the following on each form-control input component

```
handleFormControlInput = (event: any) => {
    const value = event.target.value;

    this.inputHandler(this.name, value);
    this.formControlInput.emit({ name: this.name, value: value });
  }
```

IE we don't need to be getting `name` from `event.target` since we are passing `name` as a prop I believe pretty much every time we are rendering any of these form control components

colganmeanor (approved):
Overall really great work @TiagoFuelber !

The skeleton updates look great, I think it was a good idea to add in more text parts for deeper customization. 

I left a comment showing a quirk with the radio buttons / checkbox in the Light theme in the PaymentProvisioning component. 

And the only other thing to mention is that when I pull this down and build it generates a new pnpm-lock.yaml file. I think that our local pnpm versions are different because the lockfile version on yours thats checked in says `version 9` but mine says `version 6`. 🤔 

I am not sure who needs to update their pnpm locally, me or you lol

packages/webcomponents/src/components/checkout/new-payment-method.tsx
TiagoFuelber: The click event was bubbling up and introducing some weirdness

packages/webcomponents/src/components/checkout/saved-payment-method.tsx
TiagoFuelber: It's because this component doesn't have shadow-dom, so it inherits the StyledHost styles from the parent.

packages/webcomponents/src/ui-components/form/form-control-radio.tsx
TiagoFuelber: I agree with that!

Issue #797, opened 2024-12-27
Add custom Skeletons for each component


TiagoFuelber:
Add custom loading skeletons for each component that needs has a loading state

<img width="622" alt="Image" src="https://github.com/user-attachments/assets/39205d81-bebc-4fd5-8f9e-a939b1b0fb76" />

PR #798, opened 2024-12-27
797 add custom skeletons for each component

TiagoFuelber:
Adds custom skeletons for the following components:

- Business Form
- Payment Provisioning
- Season Interruption Insurance
- Payment Details
- Payout Details

Links
-----
#797 

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------

[ ] - Each of the listed components should now display a custom skeleton when in loading state


packages/webcomponents/src/components/business-forms/business-form/business-form.tsx
```diff
@@ -106,6 +108,10 @@ export class BusinessForm {
   }
 
   render() {
+    if (this.isLoading) {
+      return <StyledHost><BusinessFormLoading /></StyledHost>;
```
colganmeanor, line 112: I'm curious why this `BusinessFormLoading` is wrapped in `StyledHost` here. Should the other custom loading components also be wrapped in Styled Host? 

If so, should we move the wrapped StyledHost to inside these components?

colganmeanor (dismissed):
Great work! 

Overall I think this is where it needs to be. I think there are definitely some small improvements that could be made to how the PaymentProvisioning component handles loading overall, but we don't need to hold this PR up for that, I think the skeletons look good. 

One small suggestion, doesn't have to be in this PR, but I like that season-interuption-insruance tests are testing a snapshot for the loading skeleton. As we proceed to add in more skeleton loading states we should be sure to update the spec tests for these components to test for the proper loading state. 

I think this PR does need a changeset so we can communicate that the loading skeleton can now be customized for each of these components using the skeleton part.

packages/webcomponents/src/components/business-forms/business-form/business-form.tsx
TiagoFuelber: Hey, in this case, `BusinessFormLoading` is wrapped in `StyledHost` because it needs the CSS injected by `StyledHost` for the loading states to render correctly. The key reason it's done here is that the loading component is conditionally rendered before the main render block of `BusinessForm`.

`StyledHost` injects CSS into the shadow DOM, typically loading global styles like Bootstrap. Since it applies styles to everything inside it, we usually wrap the topmost element to avoid duplicating this code across multiple components.

If we were to move `StyledHost` inside every loading component, it would unnecessarily duplicate the injected styles. That's why it's more efficient to load it once at a higher level.

In the payment provisioning steps, you don’t see StyledHost because the `justifi-payment-provisioning-core` component already handles loading the necessary styles. It acts as a parent wrapper, so there's no need to add `StyledHost` again.

colganmeanor: Understood - thanks for explaining this for me! That makes sense now.

colganmeanor (approved):

Issue #805, opened 2025-01-07
Optimize Github Actions to leverage caching and action dependencies


TiagoFuelber:
One of the things that inspired this is that the build runs many times (for each action) and could instead run once and be a dependency of the actions that follow, such as deploying to stroybook and publishing components, for example

PR #812, opened 2025-01-09
Optimize Github Actions to leverage caching and action dependencies 

TiagoFuelber:
This PR makes changes in the Github Actions workflows to add the cache option on the install tasks.

Reference: 
https://github.com/actions/setup-node#caching-global-packages-data
https://github.com/actions/cache
https://github.com/actions/setup-node?tab=readme-ov-file#caching-global-packages-data
https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data

Links
-----
#805 

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
[ ] - I don't think we need to directly QA this right now. But we should pay attention in next PRs and releases to the time the actions are going to take to complete.

### DIFF
--- a/.github/workflows/chromatic-deployment.yaml
+++ b/.github/workflows/chromatic-deployment.yaml
@@ -10,6 +10,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
       - uses: pnpm/action-setup@v2
         with:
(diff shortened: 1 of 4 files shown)


colganmeanor (approved):
🦖

PR #820, opened 2025-01-10
Return more accurate response data for payment provisioning mock data

### LINKED ISSUE #810: Ensure mock data is correct for PaymentProvisioning
During group QA of our v5 release we noticed that mock data for PaymentProvisioning is not aligned with real data, so our actions in Storybook are misleading as to what the component will actually return when the step completed and submit events are emitted

colganmeanor:
### Return accurate mock data for payment provisioning form's `complete-form-step-event`

This PR adjusts the mock data passthrough for patching business's to return a response object that is structured the same as a real world example. 

Also updated the mock response for the Provisioning request to include an accurate `payload` property. 

Note - this is purely for testing the component in Storybook with mock data enabled. 

Links
-----

Closes #810 

Developer QA steps
--------------------

- [x] `complete-form-step-event` returns an accurate response object at `event.detail.response`
- [x] `submit-event` includes payload property that uses our existing mock data, structured to resemble an actual payload response.


TiagoFuelber (approved):
Nice! Looks good! 💯

PR #824, opened 2025-01-14
Secondary billing form toggle for cards

### LINKED ISSUE #823: Provide the option to simplify credit card billing information
Add a `hide-card-billing-form` prop to `TokenizePaymentMethod` and `Checkout` components.
- Existing behavior is the default (show all billing address fields)
- If `hide-card-billing-form` is set to `true`, we toggle to a secondary billing form component which is slimmed down to just `postal_code` and does not have the `Billing address` title

colganmeanor:
### Secondary billing form toggle for cards

- Adds `postal-form` internal component 
- `tokenize-payment-method` and `checkout` added new boolean prop: `hideCardBillingForm` used to render `postal-form` or `billing-form`

Links
-----

Closes #823 

Developer considerations
--------------

- On any task

Developer QA steps
--------------------

You may use the updated example files for `checkout` and `tokenize-payment-method` to test the new refactors and new postal-form component. You may also use the button I added to the example files to test filling the billing form in any case presented here (`Test fill billing form`)

- [x] Test suites pass
- [x] Component library builds as expected

Checkout:

- [x] Checkout component can be completed as normal with full billing form
- [x] Checkout component can be completed with postal-form

Tokenize-payment-method:

- [x] Tokenize component can be completed as normal with full billing form
- [x] Tokenize component can be completed with postal-form


packages/webcomponents/src/components/billing-forms/billing-form-schema.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const PostalFormSchema = object({
+  address_postal_code: postalValidation.required('Enter postal code'),
+});
+
+export function isBillingFormFields(fields: any): fields is BillingFormFields {
+  return 'name' in fields && 'address_line1' in fields && 'address_city' in fields && 'address_state' in fields && 'address_postal_code' in fields;
+}
+
+export function isPostalFormFields(fields: any): fields is PostalFormFields {
```
TiagoFuelber, line 43: I suspect this function would return `true` if passing the billing form fields either. 🤔 

Aren't we looking for something like this?

```export function isPostalFormFields(fields: any): fields is { address_postal_code: string } {
  return (
    typeof fields === 'object' &&
    fields !== null &&
    Object.keys(fields).length === 1 &&
    'address_postal_code' in fields
  );
}
```

colganmeanor: Yeah, that seems much better to me. Good catch! thanks man

PR #825, opened 2025-01-15
Internal Badge UI component

jakemer:
This PR introduces a `<Badge />` UI component that can be used (internally) across all of our components that use badges. The purpose is to make sure that the css parts and classes are set in a consistent way (or at all) based on the `variant` it is given.

Links
-----

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
Ensure that tags are still displaying correctly in the following components (check between local & prod):
- `CheckoutsList`
- `PaymentsList`
- `PaymentDetails`
- `PayoutsList`
- `PayoutDetail`
- `TerminalsList`

It is also worth comparing `BusinessDetails` to production because it uses the `DetailItem` component which has changed slightly in this PR to allow rendering of the `<Badge />` component


packages/webcomponents/src/components/checkouts-list/checkouts-status.tsx
TiagoFuelber, line 1: Awesome!

TiagoFuelber (approved):
Looks good!

Issue #816, opened 2025-01-10
Update Storybook component doc pages with more examples


jakemer:
Refer to the Season Interruption docs page. It has a visual example and a "Show Code" button that will expand the code (configure this to be expanded by default if possible)

- Have examples that show the styling/css parts (maybe keep these examples simple aside from styling)
- Show _all_ event handlers and methods being used in the examples

The checkout and Tokenize Payment Methods component use methods.
 Add the usage of a method to the code example for both components

PR #839, opened 2025-01-21
816 update storybook component doc pages with more examples

TiagoFuelber:
This PR does the following:

* Adds a rendered component to the `Example usage` section on all components documentation - `Docs.mdx`.
* Improve the code examples on the docs to show also an example of styling using CSS parts.
* Improve some code examples by showing how to use the methods and events available.
* Add the `register-web-components` file to satisfy typescript about the web-components usage on the Docs files.

Links
-----
#816 

Developer QA steps
--------------------
In order to test this, Storybook should be run with `dev:mocks`.

- [X] Review all components `Docs` page. It should display a rendered component loaded with mock data
- [X] Right after the component example, it should display code examples that covers styling with CSS parts, events and methods as well.


colganmeanor (dismissed):
Overall - great work here. Really nice to see all the components in line with the doc pages. 

I'm approving - but I wanted to ask how we feel about the order of the items on the doc pages from top to bottom? 

For example - on the Season Interruption Insurance component doc page - the component example is rendered at the very top of the doc page. 

The other examples you cooked up for us here render the component examples a little further down the page in the `Example Usage` section. I'm not sure which is best, but we should make it consistent I think. Maybe let's stick with the `Example Usage` section that you went with, and let's update the insurance doc page to show that same section? 🤔

colganmeanor:
One more thing - can you make a small fix for us on each of the Doc pages here? This is something I've noticed before but haven't made time to fix - the `event usage` section should be the same header type as the other sections on the page. Right now, on each page that we have it, we render it as `##` when I was thinking we should render it as `#` so that it appears correctly in the page index, and it has the same header style as the other sections. 


Screenshot 2025-01-21 at 3 47 39 PM

TiagoFuelber:
> show that same section? 🤔

Agree, I'll move the examples a little up in the order. 👍

colganmeanor (approved):
🍔

Issue #840, opened 2025-01-21
JFC: add css part for summary


TiagoFuelber:
Customers would like to hide the `Summary` section in the checkout web component.

Add css part for `summary` so it can be hidden via css if desired.

PR #841, opened 2025-01-21
Add checkoutSummary part to checkout component and update snapshots

TiagoFuelber:
Based on a customer request, this PR adds the `checkout-summary` part, so it can be hidden.

Links
-----
#840 

Developer considerations
--------------

Developer QA steps
--------------------
 
To test this, add to `apps/component-examples/css/example.css` the following: 
```
::part(checkout-summary) { display: none; }
```

- [X] Run `pnpm dev:checkout` and the Summary section should be hidden.

### DIFF
--- a/packages/webcomponents/src/styles/parts.ts
+++ b/packages/webcomponents/src/styles/parts.ts
@@ -96,3 +96,4 @@ export const skeleton = `skeleton`;
 // This allows the billing form to be hidden, for example
 // ::part(billing-form) { display: none; }
 export const billingForm = `billing-form`;
+export const checkoutSummary = `checkout-summary`;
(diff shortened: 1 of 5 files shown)


packages/webcomponents/src/styles/parts.ts
```diff
@@ -96,3 +96,4 @@ export const skeleton = `skeleton`;
 // This allows the billing form to be hidden, for example
```
Claudia108, line 96: Would you mind updating the comment to make it more generic? Something like:
`// This allows specific sections of a component to be hidden, for example`

TiagoFuelber: Good call!

colganmeanor (approved):
🫎

PR #850, opened 2025-01-23
Update documentation to include 'Authorization' section and improve f…

TiagoFuelber:
…ormatting across multiple components

Links
-----

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------


colganmeanor (commented):
This looks great! The only thing we need to fix though is to actually make the `Authorization` link in the doc page's index work properly. 

When I placed it on the Checkout Doc page - I updated it to render the section header the same way as the others. It needs to be notated in markdown as `# Authorization` so that the index knows how it is being linked to this section. 

Maybe we remove the header from `Authorization` in `documentation-components.tsx` so that we can render the section header in markdown in each of these respective doc files. 

So for example - refactor `Authorization` to render:

```
export const Authorization = ({ actions, entity, identification }: AuthorizationProps) => {

  return (
    <>
      <p>Authorization is performed by passing a web component token as <code>auth-token</code></p>
      <ul>
        <li>
          <WebComponentTokenDescription
            actions={actions}
            entity={entity}
            identification={identification}
          />
        </li>
      </ul>
    </>
  );
}
```

And then it can be rendered in a doc file like: 

```
# Authorization

---

<Authorization
  actions={['read']}
  entity="business"
  identification="business_id"
/>
```

colganmeanor (approved):

PR #864, opened 2025-02-03
Pass through iFrameOrigin in Tokenize payment method

### LINKED ISSUE #869: Pass through optional iframe origin to tokenize component
Add iframeOrigin prop to tokenize component and pass it through to iframe-input components rendered in card-form and bank-account-form.

Allows us to specify the iframe origin used in the component inputs to correctly work in our dashboard staging environment.

colganmeanor:
### Pass through iFrameOrigin in Tokenize payment method

This PR adds an optional hidden prop to `tokenize-payment-method` allowing us to set the component up to use staging iframe origin when shown customer-admin-dashboard staging site. 

Small updates to the tokenize-payment-method example file to organize env variables. 

Links
-----

Closes #869 

Developer considerations
--------------

- On any task

Developer QA steps
--------------------

To test this locally you will need to do the following:

- Add `iframe-origin="https://components.justifi-staging.com/v2"` to `tokenize-payment-method` in the example file
- Set `component-examples` env variables to use staging info. 
- Run `pnpm install && pnpm build && pnpm dev:tokenize-payment-method`

General:

- [ ] Component library builds and runs as expected
- [x] Tests pass

Tokenize-payment-method:

- [x] Component successfully renders and loads
- [x] Component can successfully tokenize payment methods with the staging API if iframe-origin is set to `https://components.justifi-staging.com/v2`
- [x] Component can successfully tokenize payment methods with the prod API if no iframe-origin is set


apps/component-examples/examples/tokenize-payment-method.js
```diff
@@ -3,6 +3,11 @@ require('dotenv').config();
 const express = require('express');
 const app = express();
 const port = process.env.PORT || 3000;
+const authTokenEndpoint = process.env.AUTH_TOKEN_ENDPOINT || 'https://api.justifi.ai/oauth/token';
```
jakemer, line 6: Do we need the fallbacks or can we assume that there is always an env variable being set? Doesn't hurt anything but it might be cleaner to always have the `.env` file set this so that there is one place to look.

jakemer (dismissed):

apps/component-examples/examples/tokenize-payment-method.js
colganmeanor: I actually left the fallback values there as a form of psuedo documentation / example. I was thinking we should rely on the env file for dev purposes but I wanted to make sure that anyone who is looking at this file for implementation examples would be reminded that they need to hit these official endpoints for auth tokens.

TiagoFuelber: Another valid idea is to create a `.env.example` or `.env.template` and commit it.

packages/webcomponents/src/components/checkout/new-payment-method.tsx
```diff
@@ -22,6 +22,7 @@ export class NewPaymentMethod {
   @Prop() paymentMethodGroupId?: string;
   @Prop() isSelected: boolean;
   @Prop() hideCardBillingForm?: boolean;
+  @Prop() iframeOrigin?: string;
```
TiagoFuelber, line 25: Shouldn't this be not optional?

jakemer: Good point @TiagoFuelber - I think since this is passed down from components which are guaranteed to set it, it should not be optional here

colganmeanor: Good call guys - thank you!

apps/component-examples/examples/tokenize-payment-method.js
colganmeanor: Great idea @TiagoFuelber  - I added an example file!

TiagoFuelber (approved):
Good work! 💯

PR #874, opened 2025-02-06
Update bank account form schema to allow shorter bank account numbers to be entered

### LINKED ISSUE #872: PaymentProvisioning form: validation in bank account form to strict
2 customers reported not being able to finish the onboarding form because the bank data does not conform with the Payment Provisioning form.

Platform: RoasterTools
Issue: Customers unable to fill out bank account form because it requires 8 numbers but their bank account number is only 6 digits
These customers were told to add leading 0s to get around the validation for now.

To fix: allow account numbers
[From the quick research I did](https://stackoverflow.com/questions/1540285/united-states-banking-institution-account-number-regular-expression), US bank account numbers can be 4 to 17 digits long.

The backend only requires the account number to be present so updating the validation will be fine.

colganmeanor:
### Update bank account form schema to allow shorter bank account numbers to be entered

Links
-----

Closes #872 

Developer considerations
--------------

- On any task

Developer QA steps
--------------------

**General:**

- [x] Components build and run successfully
- [x] Test suites pass

Payment-provisioning component:

- [x] `account_number` field minimum character count updated to 4. Validation should only fire on submit if input is less than 4 or greater than 17.

### DIFF
--- a/.changeset/silver-masks-confess.md
+++ b/.changeset/silver-masks-confess.md
@@ -1,0 +1,5 @@
+---
+"@justifi/webcomponents": patch
+---
+
+- Updated `justifi-payment-provisioning` bank account form step. The account number field's minimum character count has been reduced to 4 to allow for shorter bank account numbers to be entered.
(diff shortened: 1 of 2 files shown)


TiagoFuelber (approved):
Sweet

PR #880, opened 2025-02-07
Checkouts list filters query param props

### LINKED ISSUE #861: Pass filters as props to Checkouts List Filters
Follow pattern established with `payments-list-filters`. 

- Implement filter props to `checkouts-list-filters`. 
- Add documentation. 
- Add CSS parts.

colganmeanor:
### Pass query params as props to `justifi-checkouts-list-filters`

- Allows query params for checkout status and payment mode to be passed as props to `checkouts-list-filters` component
- Adds new parts to adjust or hide checkouts filters menu, or specific inputs.
- Updates example files `payments-list`, `payments-list-with-filters`, `checkouts-list`, and `checkouts-list-with-filters` with organized env variables. Update sizing and display for payments-list and checkouts-list example files.

Links
-----

Closes #861 

Developer considerations
--------------

- On any task

Developer QA steps
--------------------

**General:**

- [x] App builds and runs normally
- [x] Test Suites pass

**Checkouts-List:**

For these you should run `pnpm dev:checkouts-list-with-filters` example file. 

- [x] Component builds and functions normally without passing any props to `checkouts-list-filters`
- [x] Passing in props to `checkouts-list-filters` applies corresponding query param to component's API request
- [x] Passing in props to `checkouts-list-filters` disables corresponding input in filter menu UI
- [x] Other filter UI inputs can be used in conjunction with params passed as prop. 
- [x] Clear filters button should NOT render until user manually interacts with filters UI, IE you won't see the Clear Filters button just for passing in a param as a prop
- [x] Clear filters clears out filters that were applied by the UI, but retains the ones set by props. 
- [ ] Pagination works as expected, with params supplied by props, or without it.

### DIFF
--- a/apps/component-examples/examples/checkouts-list-with-filters.js
+++ b/apps/component-examples/examples/checkouts-list-with-filters.js
@@ -3,6 +3,11 @@ require('dotenv').config();
 const express = require('express');
 const app = express();
 const port = process.env.PORT || 3000;
+const authTokenEndpoint = process.env.AUTH_TOKEN_ENDPOINT;
+const webComponentTokenEndpoint = process.env.WEB_COMPONENT_TOKEN_ENDPOINT;
+const subAccountId = process.env.SUB_ACCOUNT_ID;
+const clientId = process.env.CLIENT_ID;
+const clientSecret = process.env.CLIENT_SECRET;
 
 app.use(
   '/scripts',
@@ -12,13 +17,13 @@ app.use('/styles', express.static(__dirname + '/../css/'));
 
 async function getToken() {
   const requestBody = JSON.stringify({
-    client_id: process.env.CLIENT_ID,
-    client_secret: process.env.CLIENT_SECRET,
+    client_id: clientId,
+    client_secret: clientSecret
   });
 
   let response;
   try {
-    response = await fetch('https://api.justifi.ai/oauth/token', {
+    response = await fetch(authTokenEndpoint, {
       method: 'POST',
       headers: {
         'Content-Type': 'application/json',
@@ -34,8 +39,7 @@ async function getToken() {
 }
 
 async function getWebComponentToken(token) {
-  const response = await fetch(
-    'https://api.justifi.ai/v1/web_component_tokens',
+  const response = await fetch(webComponentTokenEndpoint,
     {
       method: 'POST',
       headers: {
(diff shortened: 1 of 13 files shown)


apps/component-examples/examples/checkouts-list-with-filters.js
```diff
@@ -71,10 +72,13 @@ app.get('/', async (req, res) => {
       <body>
         <div style="padding:25px;">
           <div>
-            <justifi-checkouts-list-filters></justifi-checkouts-list-filters>
+            <justifi-checkouts-list-filters status="completed" />
```
TiagoFuelber, line 75: I think the prop should be `checkout-status`

packages/webcomponents/src/components/checkouts-list/checkouts-list-filters.tsx
```diff
@@ -1,16 +1,32 @@
...
 import { StyledHost } from '../../ui-components';
+import { checkoutsListFilterMenu, paymentModeCheckoutsListFilterParam, statusCheckoutsListFilterParam } from '../../styles/parts';
 
 @Component({
   tag: 'justifi-checkouts-list-filters',
   shadow: true
 })
 export class CheckoutsListFilters {
+  @Prop() checkoutStatus?: ICheckoutStatus;
+  @Prop() paymentMode?: string;
```
TiagoFuelber, line 13: It would be good to have the type here, `ICheckoutPaymentMode`

TiagoFuelber:
When attempting to navigate "next" in pagination, with `payment-mode="ecom"` the following error happened. 

image

packages/webcomponents/src/components/checkouts-list/checkouts-list-filters.tsx
colganmeanor: Good call - I just pushed up a commit to address this comment. Lemme know what you think. 

I imported `ICheckoutPaymentMode` to use as the type for this prop, but then I also converted `ICheckoutStatus` to an enum in the class file, since it really doesn't have much usage outside of the class file, and could be pretty easily converted to be represented as an enum instead. 

What do you think?

TiagoFuelber: Good deal! Really like that!

TiagoFuelber (approved):
💯

Issue #833, opened 2025-01-17
Simplify Storybook Setup by Using Mock Data Only


TiagoFuelber:
**Description**:
Currently, Storybook uses the .env file to handle PROXY_API_ORIGIN for real API calls, causing issues with inconsistent auth tokens and IDs (e.g., frequent "Not Authorized" errors). To streamline development:

* Use the example-app exclusively for real API calls during development.
* Restrict Storybook to mock data only. This change allows us to:
1 - Delete apps/docs/.env.
2 - Remove logic for handling env variables in Storybook scripts.
3 - Simplify mock setup in Storybook.
4 - Reduce maintenance of base-args (e.g., auth tokens, IDs).

**Task Breakdown**
- [x] - Update `apps/docs/stories/utils/mockAllServices.ts` simplifying the logic for the mock data. (Remove the use of `__VITE_STORYBOOK_PROXY_API_ORIGIN__`)
- [x] - Update `apps/docs/.storybook/main.ts` removing the vite.js `mergeConfig` set up for `VITE_STORYBOOK_MOCKS_ENABLED`.
- [x] - Update scripts in `apps/docs/package.json` removing the use of `VITE_STORYBOOK_MOCKS_ENABLED`.
- [x] - In `.github/workflows` folder, update `storybook.yml` and `chromatic-deployment.yaml` to remove the use of `.env`.
- [x] - Update `apps/docs/globals.d.ts` removing deprecated variables.
- [x] - Delete `apps/docs/.env`
- [x] - Update `README.md` removing instructions on how to run and build with mock data and clarifying that Storybook uses mock data and the component-examples uses real data.
- [x] - Update `apps/docs/stories/utils/base-args.ts` logic removing the use of real `ids` and simplifying the logic.
- [x] - Remove `packages/webcomponents/config.ts` totally and update all files to use `proxyApiOrigin` directly from `stencil.config` -> `plugins:replace`.
- [ ] - `packages/webcomponents` and `apss/docs` should share the same `.env` file in the root folder.

For this to be successfully implemented, it is required:
- [x]  - Storybook should be deployed successfully with mock data.
- [ ] - Chromatic should be deployed successfully with mock data and `justifi-gross-payment-chart` should display data for the past 30 days only.
- [ ] - `pnpm dev` should launch Storybook

PR #884, opened 2025-02-11
833 simplify storybook setup by using mock data only

TiagoFuelber:
**Description**:
Currently, Storybook uses the .env file to handle PROXY_API_ORIGIN for real API calls, causing issues with inconsistent auth tokens and IDs (e.g., frequent "Not Authorized" errors). To streamline development:

* Use the example-app exclusively for real API calls during development.
* Restrict Storybook to mock data only. This change allows us to:

1 - Move apps/docs/.env to the root folder
2 - Simplify mock setup in Storybook.
3 - Reduce maintenance of base-args (e.g., auth tokens, IDs).

Links
-----
#833 

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------

- [x]  App builds fine, unit tests passes.

Setup: Fill up an `.env` file in the root folder as the `.env.example`. That's the only `.env` file we're gonna need right now.

- [x] Running `pnpm build && pnpm dev` should launch Storybook loaded with mock data.
- [x] Run `pnpm build && pnpm dev:[component]` should launch the example-app for any file, with the right env variables loaded.
- [x] Staging and Prod Storybook are successfully deployed and shows mock data.


.env.example
```diff
@@ -1,3 +1,5 @@
+IFRAME_ORIGIN=http://localhost:3003
+PROXY_API_ORIGIN=https://wc-proxy.justifi-staging.com
```
colganmeanor, line 2: for our example env file - do we need to include IFRAME_ORIGIN at all? Thats only going to be used by us internally. 

Also - the `PROXY_API_ORIGIN` is set to staging while the other vars in this file are showing production endpoints. Should these be the same?

packages/webcomponents/src/api/services/payment.service.ts
```diff
@@ -21,9 +20,8 @@ export class PaymentService implements IPaymentService {
     params: any,
     apiOrigin?: string
   ): Promise<IApiResponseCollection<IPayment[]>> {
-    
     if (!apiOrigin) {
-      apiOrigin = config.proxyApiOrigin;
+      apiOrigin = PROXY_API_ORIGIN;
```
colganmeanor, line 24: we can probably just remove this entire block: 

```
if (!apiOrigin) {
      apiOrigin = PROXY_API_ORIGIN;
    }
```

and update line 21 to be `apiOrigin?: string = PROXY_API_ORIGIN`

I think, right?

packages/webcomponents/src/api/services/payout.service.ts
```diff
@@ -23,9 +22,8 @@ export class PayoutService implements IPayoutService {
     params: any,
     apiOrigin?: string
   ): Promise<IApiResponseCollection<IPayout[]>> {
-
     if (!apiOrigin) {
-      apiOrigin = config.proxyApiOrigin;
+      apiOrigin = PROXY_API_ORIGIN;
```
colganmeanor, line 26: If we can remove the if/else block that sets iframe origin in the payment.service.ts file, then I guess we can do the same thing a few times in this file. 🤔

packages/webcomponents/src/api/services/terminal.service.ts
```diff
@@ -22,9 +21,8 @@ export class TerminalService implements ITerminalService {
     params: any,
     apiOrigin?: string
   ): Promise<IApiResponseCollection<ITerminal[]>> {
-    
     if (!apiOrigin) {
-      apiOrigin = config.proxyApiOrigin;
+      apiOrigin = PROXY_API_ORIGIN;
```
colganmeanor, line 25: Another one we should be able to shorten by assigning this in the interface I believe

colganmeanor (approved):
🪇

packages/webcomponents/src/api/services/payment.service.ts
TiagoFuelber: Good call! I didn't notice that at all.

.env.example
TiagoFuelber: These two variables were in the `packages/webcomponents/.env` and were the ones that Stencil was using to build the components. 

So if we don't pass the iframe-origin secret prop, the components get it from the .env file. In production, this env file is created in the CI action and it gets the iframe-origin from the Github secrets. 

In our local environment, I think we probably want to point all variables to staging.

colganmeanor (approved):

Issue #866, opened 2025-02-04
justifi-terminals-order: Display business readonly details


jakemer:
The component should:

- [x] - Display a skeleton while loading
- [x] - Check that the business had terminal orders enabled (don't do nothing for now)
- [x] - Display "Business name", "Shipping address" (business address), and contact information (Representative information)
- [x] - Handle error state and emit error event in case fetch fails
- [x] - Add CSS parts and some styling, if needed, to Storybook and example app

PR #892, opened 2025-02-14
866 justifi terminals order display business readonly details

TiagoFuelber:
This pull request adds the changes necessary for the `justifi-order-terminals` component to load readonly information about the Business, Representative and Shipping Address.

- [ ] - Creates new CSS parts for Bootstrap `list-group` and `list-group-item` classes.
- [ ] - Fetches and display the Business Name, Representative and Shipping Address information using basic bootstrap styles with parts that can be customized.
- [ ] - Display a skeleton while loading.
- [ ] - Display an error message in case of error.
- [ ] - Emits an error event in case of error.
- [ ] - Adds more unit tests to cover the new functionalities.

Links
-----
#866 

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
`pnpm dev:order-terminals`
- [ ] - It should display a loading skeleton while it's loading.
- [ ] - It should fetch and display Business, Representative and Shipping Address information.
- [ ] - [force error] - It should display the thrown error message.
- [ ] - [force error] - It should emit an error-event with details.
- [ ] - Styles can be customized using parts: `heading-4`, `text`, `background-color`, `list-group`, list-group-item`.


packages/webcomponents/src/components/order-terminals/order-terminals.tsx
```diff
@@ -69,21 +71,44 @@ export class OrderTerminals {
...
-        <StyledHost>
-          <p>Loading...</p>
-        </StyledHost>
-      );
-    }
-
     return (
       <StyledHost>
-        <p>Business Legal Name: {this.business.legal_name}</p>
+        {this.isLoading && <OrderTerminalsLoading />}
```
jakemer, line 76: Love the functional component loading state 👍

packages/webcomponents/src/components/order-terminals/order-terminals.tsx
```diff
@@ -69,21 +71,44 @@ export class OrderTerminals {
...
+                  <li class="list-group-item" part={listGroupItem}>Name: {this.business?.representative?.name}</li>
+                  <li class="list-group-item" part={listGroupItem}>Title: {this.business?.representative?.title}</li>
+                  <li class="list-group-item" part={listGroupItem}>Email: {this.business?.representative?.email}</li>
+                  <li class="list-group-item" part={listGroupItem}>Phone: {this.business?.representative?.phone}</li>
+                </ul>
+              </div>
+              <div class="col-6">
+                <h2 part={heading4}>Shipping Address:</h2>
+                <ul class="list-group" part={listGroup}>
+                  <li class="list-group-item" part={listGroupItem}>Street: {this.business.legal_address?.line1 || ''}</li>
```
jakemer, line 101: I think the optional chaining will return `undefined` if there is no value, which should not show up on the page. So do we need the `|| ''` condition?

jakemer (dismissed):

packages/webcomponents/src/components/order-terminals/order-terminals.tsx
TiagoFuelber: @jakemer Good catch, I removed the optional chaining and it stll worked fine!

jakemer (approved):

PR #895, opened 2025-02-17
Create terminal orders list component

### LINKED ISSUE #887: justifi-terminal-orders-list: create component and example file
First steps that can be done before additional details are finalized:

- Create component / core component that can will eventually render table component
- Set up example file to use for developing, testing, and QA

colganmeanor:
### Create terminal orders list component

- Creates `justifi-terminal-orders-list` component, `terminal-orders-list-core` component, `terminal-orders-table` logic, and `terminal-orders-list` js example file. 
  - there is no data being fetched yet as the API is not ready, so API service files will come later
  - columns are placeholder values for now until we decide how the data should be rendered and organized
  - classes and types are `any` for now until we have a firm idea of what the shape of the data is
  - tests will come in a later PR

Links
-----

Closes #887 

Developer considerations
--------------

- On any task

Developer QA steps
--------------------

- [x] Component library builds and runs

To run the new component's example file: `pnpm dev:terminal-orders-list`

- [x] Table component renders with `No Results` / empty state displayed

### DIFF
--- a/apps/component-examples/examples/terminal-orders-list.js
+++ b/apps/component-examples/examples/terminal-orders-list.js
@@ -1,0 +1,89 @@
+require('dotenv').config({ path: '../../.env' });
+
+const express = require('express');
+const app = express();
+const port = process.env.PORT || 3000;
+// const authTokenEndpoint = process.env.AUTH_TOKEN_ENDPOINT;
+// const webComponentTokenEndpoint = process.env.WEB_COMPONENT_TOKEN_ENDPOINT;
+// const subAccountId = process.env.SUB_ACCOUNT_ID;
+const businessId = process.env.BUSINESS_ID;
+// const clientId = process.env.CLIENT_ID;
+// const clientSecret = process.env.CLIENT_SECRET;
+
+app.use(
+  '/scripts',
+  express.static(__dirname + '/../node_modules/@justifi/webcomponents/dist/')
+);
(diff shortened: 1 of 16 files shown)


apps/component-examples/examples/terminal-orders-list.js
```diff
@@ -0,0 +1,89 @@
...
+    <!DOCTYPE html>
+    <html>
+      <head>
+        <title>JustiFi Terminal Orders List Component</title>
+        <script type="module" src="/scripts/webcomponents/webcomponents.esm.js"></script>
+        <link rel="stylesheet" href="/styles/theme.css">
+        <link rel="stylesheet" href="/styles/example.css">
+      </head>
+      <body>
+        <div style="padding:25px;">
```
jakemer, line 74: We're probably going to want padding on other components (if we don't already) so maybe it makes sense to create a `component-example-wrapper` class and put it in the `example.css`?

jakemer (dismissed):
Nice. Looks like a good start - nice to get the boilerplate out of the way so that subsequent PRs are smaller

apps/component-examples/examples/terminal-orders-list.js
colganmeanor: When we discussed this in standup this morning I said I would do it in another PR, but I just decided to go ahead and include this in this PR. All the list component example files use this same class now that I added to `example.css`: `.list-wrapper`

Issue #867, opened 2025-02-04
justifi-terminals-order: Select terminal device quantities


jakemer:
Add the UI for selecting terminal devices/device quantities for the order. We will need to fetch the terminal models/images. It should include:
- Device model name
- Device image
- Quantity selector
- "Learn More" link uses description url
- Limit the orders to available number of devices (this will be sent from the backend)

<img width="707" alt="Image" src="https://github.com/user-attachments/assets/c6bddbfa-2091-4dea-905f-8124fe5d362d" />

PR #896, opened 2025-02-17
867 justifi terminals order select terminal device quantities

TiagoFuelber:
This PR commits the following

terminals-quantity-selector-skeletons

image

Links
-----
#867 

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
Run `pnpm build && pnpm dev:order-terminals`
- [ ] - Should display skeletons for the terminal-quantity-selector parts.
- [ ] - Should display an error message if fetching terminal models fail.
- [ ] - An error event should be logged if it fails to fetch terminal models.

Set a mock response: 
- Go to `mockData/mockTerminalModels.json` and copy it's content.
- Go to the network tab, right click the following request: `https://wc-proxy.justifi-staging.com/v1/terminals/order_models`,
- Click the `Override content` option.
- Paste the copied JSON and save. It should be sufficient to mock the request from now on. 

- [ ] - Terminal models should load and display accordingly with the mock data.
- [ ] - It should be possible to click "-" and "+" to add and remove terminals.
- [ ] - The sum of all added terminals should not pass `order_limit` that comes from `terminals/order_models` request.
- [ ] - It should be possible to add styles using the parts in the HTML structure


packages/webcomponents/src/api/TerminalModel.ts
```diff
@@ -0,0 +1,36 @@
+export interface iTerminalModel {
```
jakemer, line 1: For consistency sake, I think we usually name interfaces with a capital `I`, such as `ITerminalModel`

packages/webcomponents/src/api/TerminalModel.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export interface iTerminalModelsApiResponse {
+  terminal_models: iTerminalModelApiResponse[];
+  order_limit: number;
+}
+
+export class TerminalModel implements iTerminalModel {
+  public id: string;
+  public modelName: string;
```
jakemer, line 24: Should this be snake case along with `imageUrl` and `helpUrl`?

packages/webcomponents/src/api/TerminalOrder.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  constructor(terminalOrder?: ITerminalOrder) {
+    this.businessId = terminalOrder?.businessId || '';
+    this.accountId = terminalOrder?.accountId || '';
+    this.provider = terminalOrder?.provider || '';
+    this.orderType = terminalOrder?.orderType || '';
+    this.orderItems = terminalOrder?.orderItems || [];
+  }
+
+  public get payload() {
```
jakemer, line 29: Can we avoid this if we just keep the naming convention 1:1 with the backend? Keep it all snake case? Having this getter is helpful but I think going this way is prone to confusion and naming typos

packages/webcomponents/src/components/order-terminals/order-terminals.tsx
```diff
@@ -63,27 +79,101 @@ export class OrderTerminals {
...
+                  <li class="list-group-item" part={listGroupItem}>Name: {this.business?.representative?.name}</li>
+                  <li class="list-group-item" part={listGroupItem}>Title: {this.business?.representative?.title}</li>
+                  <li class="list-group-item" part={listGroupItem}>Email: {this.business?.representative?.email}</li>
+                  <li class="list-group-item" part={listGroupItem}>Phone: {this.business?.representative?.phone}</li>
+                </ul>
+              </div>
+              <div class="col-6">
+                <h2 part={heading4}>Shipping Address:</h2>
+                <ul class="list-group" part={listGroup}>
+                  <li class="list-group-item" part={listGroupItem}>Street: {this.business.legal_address?.line1 || ''}</li>
```
jakemer, line 146: Is the `|| ''` condition necessary? I think with optional chaining it'll either have a value or be `undefined` which should not show up on the page

packages/webcomponents/src/components/order-terminals/order-terminals.tsx
```diff
@@ -63,27 +79,101 @@ export class OrderTerminals {
...
+                </ul>
+
+              </div>
+            </div>
+          </div>
+        )}
+
+        {this.terminalsIsLoading && (
+          <div class="mt-5">
+            <TerminalSelectorLoading />
```
jakemer, line 160: Love the loading state functional components 👍

jakemer (dismissed):

packages/webcomponents/src/api/TerminalModel.ts
TiagoFuelber: Good catch!

packages/webcomponents/src/api/TerminalModel.ts
TiagoFuelber: Good call! Sorry, I didn't look at the other model files. They are all snake case. Changing that 👍

jakemer (approved):

PR #907, opened 2025-02-21
Improve CSV Download on Payouts List

### LINKED ISSUE #899: Clicking on CSV download in payouts-list emits same click event as table row click
**Describe the bug**

As a user implementing the payouts-list component, if you are using the click-event to navigate to a details page, there is no way to discern clicking on the CSV download link vs the rest of the table row. If a user clicks on the CSV link it will register as a row click event and send out the click event with the tableRow name. 

**To Reproduce**

1. Run `payouts-list` example file locally using `pnpm dev:payouts-list`
2. Pick a table row and click on the CSV link at the far right end of the table.
3. Clicking on the CSV link emits the click event with the name rowClick which could interfere with implementations that use this event to manage loading detail screens or views (the way we do in our customer-admin-dashboard!)

### LINKED ISSUE #903: Bug: payouts-list component CSV download is broken
**Describe the bug**
Clicking the CSV download link for a given payout table row in payouts-list seems to initiate an incorrectly formed request. 

**To Reproduce**

1. Use the example file to run a local example: `pnpm dev:payouts-list`
2. On any table row that loads, click on the CSV download option. 
3. The CSV will not be downloaded, and the network request will read the request URL as something like: `https://wc-proxy.justifi-staging.com/v1/reports/payouts/[object%20Object]`

**Expected behavior**
The CSV file should download to the users computer when this link / button is clicked by the user. 

**Screenshot**
Image

**Screen Recording**

https://github.com/user-attachments/assets/1504925b-b4c7-4709-a196-1153ecae3ec5

colganmeanor:
### Improve CSV Download on Payouts List

- Prevents table row click event from being emitted when user clicks on the CSV Download link
- Fixes cases where CSV download link was incorrectly formed, leading to failed download requests.

Links
-----

Closes #903 
Closes #899 

Developer considerations
--------------

- On any task

Developer QA steps
--------------------

General:

- [ ] Components build and run
- [ ] Test suites pass

`payouts-list`:

- [ ] Component builds
- [ ] Clicking on CSV download link correctly triggers download request to browser
- [ ] Clicking on CSV download link does NOT emit click event
- [ ] Clicking elsewhere on the table row still emits click event as expected.

### DIFF
--- a/.changeset/ten-geckos-jump.md
+++ b/.changeset/ten-geckos-jump.md
@@ -1,0 +1,6 @@
+---
+"@justifi/webcomponents": patch
+---
+
+- Fixed bug in `justifi-payouts-list` where downloading the payout CSV could return an error from a bad request. 
+- Updated behavior of `click-event` in `justifi-payouts-list` to not emit when clicking on the CSV download link.
--- a/packages/webcomponents/src/api/Payout.ts
+++ b/packages/webcomponents/src/api/Payout.ts
@@ -91,10 +91,6 @@ export class Payout implements IPayout {
     this.updated_at = payout.updated_at;
   }
 
-  get csv(): string {
-    return this.id;
-  }
-
   formattedPaymentAmount(amount: number): string {
     return formatCurrency(amount, this.currency);
   }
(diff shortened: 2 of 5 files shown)


jakemer (approved):
Perfect! Nice catch on this!

Issue #919, opened 2025-02-26
Improve `terminal-quantity-selector` component to use card bootstrap styles


TiagoFuelber:
The "Shipping Address" would only show if the shipping prop is set to true

So there's the styling part, and then the adding the shipping prop. Probably the styling/markup can be done last but I'll leave that up to you

The shipping prop will also set the order_type to boarding_only or boarding_shipping on the Order payload when submitting

And then last thing... can we put the terminal quantity selector and other sub-components into the order-terminals folder? I would like to get to a point where the components folder top-level items are only exported components, nothing internal (to keep it clean)

Issue #918, opened 2025-02-26
Indicate Max Quantity to Users


TiagoFuelber:
**Description**
Currently, there is no clear way for users to see the max quantity allowed. We should add an indicator to improve clarity.

Displaying the max quantity above the list of terminal models.

Issue #868, opened 2025-02-04
justifi-terminals-order: Submit terminal order


jakemer:
- Submit the terminal order
- Emit an event when the order request has succeeded/failed (can be used for re-routing and confirmation/error messages)
- Re-enable Story [ff0e3d8cf6296d17752bc23b0d0158feb7a4be45](https://github.com/justifi-tech/web-component-library/commit/ff0e3d8cf6296d17752bc23b0d0158feb7a4be45)

PR #925, opened 2025-02-27
868 justifi terminals order submit terminal order

TiagoFuelber:
**Description**

- Implemented the submission logic for the `justifi-order-terminals` component.
- Added a new prop `shipping` which sets the `order_type` and show/hide the shipping address details from the page.
- Added a tooltip to show the Terminal model description.
- Added a "Order limit" indicator below the model cards.
- Small layout adjustments.

Links
-----
#868 
#919 
#918 

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------

- [x] - "Submit Order" button should be disabled while no terminal is added.
- [x] - It should NOT be possible to add more terminals than the quantity indicated by the "Order limit".
- [x] - If `shipping` prop is passed with "true", it should display the shipping address details below the Terminal Model Selectors.
- [x] - If `shipping` prop is true, the order payload should have `order_type: 'boarding_shipping'`.
- [x] - If `shipping` prop is not set, the order payload should have `order_type: 'boarding_only'`. 
- [x] - An order can be successfully created.

### DIFF
--- a/packages/webcomponents/src/components/order-terminals/order-terminals.tsx
+++ b/packages/webcomponents/src/components/order-terminals/order-terminals.tsx
@@ -34,7 +34,6 @@ export class OrderTerminals {
   @State() orderLimit: number;
   @State() order: TerminalOrder;
   @State() totalQuantity: number = 0;
-  @State() wasClicked: string | boolean = 'no';
 
   analytics: JustifiAnalytics;
 
--- a/packages/webcomponents/src/components/order-terminals/order-terminals.tsx
+++ b/packages/webcomponents/src/components/order-terminals/order-terminals.tsx
@@ -9,8 +9,8 @@ import { TerminalModel, ComponentErrorEvent } from "../../api";
 import { Business } from "../../api/Business";
 import { makeGetBusiness } from "../../actions/business/get-business";
 import { OrderTerminalsLoading } from "./order-terminals-loading";
-import { buttonPrimary, heading4, inputCheckbox, listGroup, listGroupItem, text } from "../../styles/parts";
-import { TerminalSelectorLoading } from "../terminal-quantity-selector/terminal-quantity-selector-loading";
+import { buttonPrimary, heading4, heading5, text } from "../../styles/parts";
+import { TerminalSelectorLoading } from "./terminal-quantity-selector/terminal-quantity-selector-loading";
 import { makeGetTerminalModels } from "../../actions/terminal/get-terminal-models";
 import { TerminalService } from "../../api/services/terminal.service";
 import { TerminalOrder, TerminalOrderType } from "../../api/TerminalOrder";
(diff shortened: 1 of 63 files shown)


packages/webcomponents/src/components/order-terminals/order-terminals.tsx
```diff
@@ -23,23 +24,32 @@ export class OrderTerminals {
...
   @State() errorMessage: string;
   @State() businessIsLoading: boolean = true;
   @State() terminalsIsLoading: boolean = true;
   @State() business: Business;
   @State() terminalModels: TerminalModel[];
   @State() orderLimit: number;
-  @State() order: TerminalOrder = new TerminalOrder({});
+  @State() order: TerminalOrder;
   @State() totalQuantity: number = 0;
+  @State() wasClicked: string | boolean = 'no';
```
colganmeanor, line 37: Is `wasClicked` being used anywhere in this file? I cant find any other usage, just curious what this would be for

packages/webcomponents/src/components/order-terminals/order-terminals.tsx
```diff
@@ -119,61 +152,79 @@ export class OrderTerminals {
...
+                  limit={this.orderLimit - this.order.totalQuantity}
+                  onSelectedQuantityChange={this.onSelectedQuantityChange.bind(this)}
+                />
+              ))}
+            </div>
+          )}
+
+          <div class="mt-3 text-end">Order limit: {this.orderLimit}</div>
+
+          {!this.businessIsLoading && this.business && this.shipping && (
```
colganmeanor, line 204: This is acceptable for right now - but we should consider how we can clean up these conditional loading states. Maybe using different render states to define when things like `this.business` and `this.terminalModels` are both finished loading. 

Like I said, fine for now because I know we wanna the MVP out the door, but something to think about when we have time to clean this up a little more! 🌟

colganmeanor (approved):
Everything makes sense and looks good! 

I know we are focused on getting this out the door ASAP, but when we have the time we should also consider implementing the core component pattern here that we have implemented on our other components. 

Great work overall! Really exciting to have this so close to being released. 🙏🏻

colganmeanor:
Approved Dev QA 🔮

I noticed it's possible to keep clicking the submit button and creating orders when the component is up and running. Definitely something to look at. Are there plans to disable the submit button or put the component into any kind of `complete / submitted` state once an order has been posted?

packages/webcomponents/src/components/order-terminals/order-terminals.tsx
TiagoFuelber: Hey man! That's weird, I really don't remember adding this. And for sure it's not being used, thanks for pointing that! Removing.

TiagoFuelber:
> Approved Dev QA 🔮
> 
> I noticed it's possible to keep clicking the submit button and creating orders when the component is up and running. Definitely something to look at. Are there plans to disable the submit button or put the component into any kind of `complete / submitted` state once an order has been posted?

Yeah, good point! 

I added now a `submitting` state so we can disable the button and indicate it while the request is happening. 👍

colganmeanor (approved):

PR #944, opened 2025-03-12
Refactor monetary inputs for use outside of justifi-payment-provisioning

jakemer:
The `form-control-monetary` field has only been used in the payment provisioning form so far, where amounts are submitted as whole numbers (dollars).

Elsewhere (within our Payments system), amounts are submitted in the smallest currency denomination (cents) and need to be formatted differently.

Example:
In the provisioning form: user inputs "1,000", which gets submitted as "1000" (dollars)
In other places (such as the refund component in this PR): user inputs "1,000.00" which gets submitted as "100000" (cents)

We had been using `imask` for the `form-control-monetary` before, which really does not play nicely with the with the second example. Fortunately, achieving the behavior we want isn't too difficult to do manually.

Changes:
- Move `form-control-monetary` to the provisioning component directory and rename it to `form-control-monetary-provisioning` to prevent confusion. It should only be used here anyhow.
- Make a new component named `form-control-monetary` to act as our primary currency input (cents shown as decimal dollar value)
- Internally manage form vs display values in that component
- Refactor `formatCurrency` helper to also work for the new input (omit the currency symbol) so that the code is shared and consistent
- Test the `formatCurrency` helper changes to ensure no regressions

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------

New monetary input:
- [x] Pull this branch and run `pnpm install`
- [x] Run `pnpm dev:refund-payment`, and go to `localhost:3420`
- [x] The amount field should default to the value passed in the `refund-payment.js` file (`payment-amount-refundable` prop)
- [x] In the "amount" field, try entering different values and characters. Only numbers should be accepted, and the numbers should be formatted as expected (thousands separator and two decimal places, for example 1,234.56)
- [x] Trigger validation. Any entered value that exceeds the `payment-amount-refundable` should throw an error saying that it cannot exceed that amount (`$10.00`)
- [x] The submit button text should update to the entered amount saying "Refund $x.yz"

Provisioning monetary input:
- [x] Run `pnpm dev:payment-provisioning`
- [x] Proceed through the form until you arrive at "Additional Questions" and enter some values into the `$` fields. These fields should not display any decimals, but should still have commas as thousand separators


packages/webcomponents/src/components/refund-payment/refund-payment-schema.ts
```diff
@@ -0,0 +1,24 @@
...
+const RefundPaymentSchema = (maxRefundAmount: string) => yup.object().shape({
+  amount: yup
+    .string()
+    .required('Amount is required')
+    .typeError('Amount must be a number')
+    .test(
+      'amount',
+      `Refund amount cannot be more than ${formatCurrency(+maxRefundAmount)}`,
+      function (value) {
+        return value <= this.options.context.amount;
```
TiagoFuelber, line 18: While testing this line was throwing an error on the console. `this.options.context.amount` was `undefined`.

Wonder if this should be `+value <= +maxRefundAmount`? 

Will this be changed in the upcoming refund-component PR?

packages/webcomponents/src/components/refund-payment/refund-payment.tsx
```diff
@@ -82,21 +72,18 @@ export class RefundForm {
   async handleSubmit(event: Event) {
     event.preventDefault();
 
-    this.formController.validateAndSubmit(this.submitRefund.bind(this), {
-      amount: this.amount,
-    });
+    this.formController.validateAndSubmit(this.submitRefund);
```
TiagoFuelber, line 75: I saw some console errors like `not able to set isSubmitting for undefined`. 
Wonder if `this.submitRefund` should `bind(this)` ??

Not sure if this should be a problem or not, considering QA plan passed.

packages/webcomponents/src/components/refund-payment/refund-payment-schema.ts
jakemer: Oh yes, good point. I forgot to update `this.options.context.amount` when I updated this to a function that takes `maxRefundAmount`. Will fix this, thanks!

packages/webcomponents/src/components/refund-payment/refund-payment.tsx
jakemer: I maybe shouldn't have done anything with the submit method in this PR, but submission should be handled in a subsequent PR so let's worry about it then. We should follow the patterns of other components that use forms.

TiagoFuelber (dismissed):
💯

TiagoFuelber (approved):
💯

Issue #968, opened 2025-03-25
justifi-order-terminals: Add "buttonLabel" prop and redo Address info logic


TiagoFuelber:
- Add a `submitButtonText` label. 
- Change the Address logic to display the text based on the `shipping` prop. 
- - `shipping=true`: "Shipping Address".
- - `shipping=false`: "Business Address".

PR #969, opened 2025-03-25
Add customizable submit button text and improve address header display in OrderTerminals component

TiagoFuelber:
**This PR commits the following**

- [ ] - Add the `submitButtonText` prop.
- [ ] - Change the Address info header text based on the `shipping` prop value. 

Links
-----
#968 

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
Run `pnpm dev:order-terminals`.

- [x] - If `submitButtonText` is not passed, the submit button text should be "Submit Order".
- [x] - If `submitButtontext` is preset, it should change the button text.
- [x] - `shipping="true"`: Address info header should be "Shipping Address".
- [x] - `shipping="false"`: Address info header should be "Business Address".

### DIFF
--- a/packages/webcomponents/src/components/order-terminals/order-terminals.tsx
+++ b/packages/webcomponents/src/components/order-terminals/order-terminals.tsx
@@ -40,7 +40,7 @@ export class OrderTerminals {
   @State() order: TerminalOrder;
   @State() totalQuantity: number = 0;
 
-  addressHeader = this.shipping ? 'Shipping Address: ' : "Business Address: ";
+  addressHeader = this.shipping ? "Shipping Address:" : "Business Address:";
 
   analytics: JustifiAnalytics;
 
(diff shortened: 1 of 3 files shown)


packages/webcomponents/src/components/order-terminals/order-terminals.tsx
```diff
@@ -39,6 +40,8 @@ export class OrderTerminals {
   @State() order: TerminalOrder;
   @State() totalQuantity: number = 0;
 
+  addressHeader = this.shipping ? 'Shipping Address: ' : "Business Address: ";
```
jakemer, line 43: Can we make the quotes consistent (single vs double) and also I think we can remove the space at the end of each heading

jakemer (approved):

PR #121, opened 2023-06-29
FE-150 deprecate payment method component props

TiagoFuelber:
* Show deprecation warning messages when the user listen to events that will be deprecated:
-- cardFormTokenize 
-- bankAccountFormTokenize
-- cardFormValidate
-- bankAccountFormValidate
-- cardFormReady
-- bankAccountFormReady

* Add a "ready" event to replace cardFormReady/bankAccountFormReady in next release.
* Add a new section to the README advising about the deprecation

Links
-----

Developer considerations
--------------

- [ ] Increment the version number for relevant packages
  - [ ] This is a new MAJOR version (incompatible API changes were made)
  - [ ] This is a new MINOR version (functionality was added in a backwards compatible manner)
  - [ ] This is a new PATCH version (backwards compatible bug fixes)
- [ ] Update the CHANGELOG.md with a summary of the changes
  - If this is a MAJOR version, what are the breaking API changes? Was anything added to the API?
  - If this is a MINOR version, what feature was added? Why would consumers want this new version?
  - If this is a PATCH version, what was fixed?
- [ ] Update the appropriate documentation. Keep in mind that this mono repository has README files at the repository, project, and component levels. We want developers to be able to find what they need from the main README, so make sure to add links to component documentation if adding something new.

Steps for Testing/QA
--------------------

PR #988, opened 2025-04-02
refactor: update checkout API endpoint and improve error handling

TiagoFuelber:
This PR commits the following fixes: 

- Checkout example: `http://` on the `origin_url` was breaking the request.
- Checkout with insurance example: Same `origin_url` issue and switched from the proxy API to the regular one.

Links
-----

#121 

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------

- [ ] `pnpm dev:checkout` should work fine.
- [ ] `pnpm dev:checkout-with-insurance` should work fine.

### DIFF
--- a/apps/component-examples/examples/checkout-with-insurance.js
+++ b/apps/component-examples/examples/checkout-with-insurance.js
@@ -3,7 +3,7 @@ require('dotenv').config({ path: '../../.env' });
 const express = require('express');
 const app = express();
 const port = process.env.PORT || 3000;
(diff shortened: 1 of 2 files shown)


colganmeanor (approved):
Looks good! Thanks for investigating this one man

One small request would be to update the insurance policy values to use `2025` instead of `2024` so that the quotes request doesn't fail when the example is loading. But no big deal either way!

PR #989, opened 2025-04-02
Refund Component: Get Payment and Post Refund

colganmeanor:
### Refund Component: Get Payment and Post Refund

- Add Refund API service and actions files to handle API requests
  - adds new prop: `accountId` to pass as header in refund request. 
- Add `Refund.ts` type file
- Adjust schema for refund component form
-  Refactor and improve form controller loading, subscribing to values and errors
- Pass `final` block and `apiOrigin` props to payment detail actions and services
- Adjust Refund Example file with new props and log submit-event

Links
-----

Closes https://github.com/justifi-tech/engineering-artifacts/issues/63

Developer considerations
--------------

- On any task

Developer QA steps
--------------------

**General:**

- [x] Component library builds and runs without errors
- [x] Tests pass

**Refund Component:**

In order to properly test the component, you will need a valid payment ID to supply as a prop to the component. 

- [x] Component successfully loads payment data and uses `payment.amount_refundable` to generate form values
- [x] If payment.amount_refundable === 0 then an `error-event` is emitted from the component with `ComponentErrorCodes.INVALID_REFUND_AMOUNT` (an easy way to test this would be to post a full refund for a valid payment, and then refresh the browser page to reload the component with the same payment ID)
- [x] Form validation is triggered if attempting to post a refund with manually entered `0.00` value
- [x] Form validation is triggered if attempting to post a refund with a manually entered value that is GREATER than the amount_refundable from the payment. 
- [x] Refund can be successfully posted with the specified amount, reason, and description from the form
- [x] Successful post emits submit-event with successful refund API response
- [x] Unsuccessful post emits submit-event with error response and emits error-event


TiagoFuelber (approved):
💯

PR #994, opened 2025-04-04
Add CSS part for $ input adornment

colganmeanor:
### Add CSS part for $ input adornment

- Adds new part for input adornment shown with `form-control-monetary`
- Updates theme.css to add simple styling to the input adornment 

Links
-----

Closes https://github.com/justifi-tech/engineering-artifacts/issues/142

Developer QA steps
--------------------

- [x] Component library builds and runs
- [x] Test suites pass
- [x] Input adornment on `form-control-monetary` can be styled by targeting the new CSS part `input-group-text`(this is the name of the bootstrap class used to style the adornment from the official bootstrap docs) (this can be tested fairly simply by rendering the refund-payment.js example file and adjusting `theme.ts`)
- [x] **Light** theme on storybook successfully targets and styles input adornment (you may quickly check this by looking at the docs example for `BusinessForm` component)
- [x] **Dark** theme on storybook successfully targets and styles input adornment (you may quickly check this by looking at the docs example for `BusinessForm` component)

### DIFF
--- a/packages/webcomponents/src/styles/parts.ts
+++ b/packages/webcomponents/src/styles/parts.ts
@@ -35,12 +35,6 @@ export const inputDisabled = `input-disabled ${input}`;
 export const inputFocused = `input-focused ${input}`;
 export const inputInvalidAndFocused = `input-invalid-and-focused ${input}`;
 
-export const inputAdornment = `input-group-text ${text}`;
-export const inputAdornmentInvalid = `input-invalid ${inputAdornment}`;
-export const inputAdornmentDisabled = `input-disabled ${inputAdornment}`;
-export const inputAdornmentFocused = `input-focused ${inputAdornment}`;
-export const inputAdornmentInvalidAndFocused = `input-invalid-and-focused ${inputAdornment}`;
-
 export const inputRadio = `input-radio`;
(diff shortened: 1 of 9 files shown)


packages/webcomponents/src/styles/parts.ts
```diff
@@ -47,6 +47,8 @@ export const inputCheckboxFocused = `input-checkbox-focused ${inputCheckbox}`;
 export const inputCheckboxChecked = `input-checkbox-checked ${inputCheckbox}`;
 export const inputCheckboxCheckedFocused = `input-checkbox-checked-focused ${inputCheckbox}`;
 
+export const inputAdornment = `input-group-text`;
```
TiagoFuelber, line 50: I’d recommend two small changes for clarity and consistency:

Rename the part string to "input-adornment" – This keeps it consistent with the variable name and the applied string. While using Bootstrap class names is usually a good guideline, in this case, input-adornment feels more descriptive of its purpose.

Add the `text` part to the composition – This ensures the element inherits the correct color and font-family styles.

colganmeanor: Good call on both suggestions! Thanks @TiagoFuelber 😄

PR #1006, opened 2025-04-14
Add JustifiCardForm and CardFormSkeleton components

TiagoFuelber:
**This PR includes the following changes:**

* Adds the new `justifi-card-form` component to the `modular-checkout-parts` folder.
* Copies `card-form-skeleton` into the `modular-checkout-parts` folder.
* Updates `checkout-wrapper` → `tokenizePaymentMethod`: adds an `addressPostalCode` argument to handle cases where the user doesn’t include the `postal-code` component.

**Key differences between the new CardForm and the old one:**

* The new version uses Shadow DOM, `StyledHost`, and `hidden-input` to support styling, as the checkout wrapper doesn’t use Shadow DOM.
* It sets a default value for the `iframeOrigin` prop.
* It initializes `checkPkgVersion` and `analytics`.

Links
-----
https://github.com/justifi-tech/engineering-artifacts/issues/100

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
- Run `pnpm dev:modular-checkout`.

- [x] - You should see the `card-form` rendered and styles are good.
- [x] - The card form should show validation errors if attempted to submit without filling it.
- [x] - Fill it correctly, submit it, you should see the successful request in the network tab.

### DIFF
--- a/packages/webcomponents/src/components/modular-checkout-parts/card-form.tsx
+++ b/packages/webcomponents/src/components/modular-checkout-parts/card-form.tsx
@@ -1,0 +1,122 @@
+import { Component, h, Method, Prop, State } from "@stencil/core";
+import CardFormSkeleton from "./card-form-skeleton";
+import { StyledHost } from "../../ui-components";
+import JustifiAnalytics from "../../api/Analytics";
+import { checkPkgVersion } from "../../utils/check-pkg-version";
+
+@Component({
+  tag: "justifi-card-form",
+  shadow: true,
+})
+export class JustifiCardForm {
+  @Prop() iframeOrigin?: string = IFRAME_ORIGIN;
+
+  @State() isReady: boolean = false;
+
+  private cardNumberIframeElement!: HTMLIframeInputElement;
+  private expirationMonthIframeElement!: HTMLIframeInputElement;
+  private expirationYearIframeElement!: HTMLIframeInputElement;
+  private cvvIframeElement!: HTMLIframeInputElement;
+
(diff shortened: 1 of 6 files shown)


packages/webcomponents/src/components/modular-checkout-parts/card-form.tsx
```diff
@@ -0,0 +1,122 @@
...
+  @Prop() iframeOrigin?: string = IFRAME_ORIGIN;
+
+  @State() isReady: boolean = false;
+
+  private cardNumberIframeElement!: HTMLIframeInputElement;
+  private expirationMonthIframeElement!: HTMLIframeInputElement;
+  private expirationYearIframeElement!: HTMLIframeInputElement;
+  private cvvIframeElement!: HTMLIframeInputElement;
+
+  analytics: JustifiAnalytics;
```
colganmeanor, line 21: should analytics be in both this and the checkout-wrapper component? Feels like maybe only the top level component should handle analytics, what do you think?

colganmeanor (approved):
🍝

packages/webcomponents/src/components/modular-checkout-parts/card-form.tsx
TiagoFuelber: @colganmeanor 
I am not sure either, but I think it might be helpful to have it in both. This way we can track exactly which components the customer is slotting in the checkout wrapper.

PR #1021, opened 2025-04-30
Add optional `account-id` prop to `payout-transactions-list`

colganmeanor:
### Add optional `account-id` prop to `payout-transactions-list`

- Adds optional `account-id` prop to `payout-transactions-list` web component. The component successfully auths and requests data with a standard WCT token, but will fail the request if using the dashboard token without passing `sub-account` header to the request. This PR adds in an optional `account-id` prop for us to use in the dashboard. This prop will not need to be documented as our regular users should not run into this issue and wouldn't need to pass in this prop. 

Links
-----

Closes https://github.com/justifi-tech/engineering-artifacts/issues/287

Developer QA steps
--------------------

- [x] Component successfully requests and loads data in example file with no changes. Just passing `payout-id` and `auth-token` props

The next item to test requires us to emulate the dashboard setup a little bit. We need to pass in the auth token from the dashboard, and then edit the props in the example file to pass in the token and an account ID.

Grab an auth token from the dashboard staging (either locally, or on the deployed staging site). In the `token` response, copy and save the value for `id_token` and use it for the `auth-token` value in the component props. 

Additionally - add the following attribute to pass in the account:  `account-id="${subAccountId}"`

- [x] Using the dashboard auth token, and the account ID prop - the component should successfully fetch data in the example file. `sub-account` header shows in the network console and correctly sends the account id passed in the component props.

### DIFF
--- a/packages/webcomponents/src/actions/payout/get-payout-transactions.ts
+++ b/packages/webcomponents/src/actions/payout/get-payout-transactions.ts
@@ -3,10 +3,10 @@ import { ComponentErrorSeverity } from '../../api/ComponentError';
 import { getErrorCode, getErrorMessage } from '../../api/services/utils';
 
 export const makeGetPayoutTransactions =
-  ({  authToken, service, apiOrigin }) =>
+  ({  authToken, accountId, service, apiOrigin }) =>
   async ({ params, onSuccess, onError, final }) => {
     try {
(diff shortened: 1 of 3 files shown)


jakemer (approved):
Makes sense and looks good 👍

PR #1022, opened 2025-04-30
99 refactor export justifi billing info component

TiagoFuelber:
Links
-----

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------

### DIFF
--- a/packages/webcomponents/src/components/checkout-wrapper/checkout-wrapper.tsx
+++ b/packages/webcomponents/src/components/checkout-wrapper/checkout-wrapper.tsx
@@ -12,8 +12,8 @@ import { BillingInfo } from "../../api/BillingInformation";
 export class CheckoutWrapper {
   analytics: JustifiAnalytics;
   private observer?: MutationObserver;
-  private cardFormRef?: HTMLJustifiCardFormElement;
-  private bankAccountFormRef?: HTMLJustifiBankAccountFormElement;
+  private paymentMethodFormRef?: HTMLJustifiCardFormElement | HTMLJustifiBankAccountFormElement;
+  private billingInfomationFormRef?: HTMLJustifiBillingInformationFormElement | HTMLJustifiPostalCodeFormElement;
 
   @Prop() authToken: string;
   @Prop() accountId: string;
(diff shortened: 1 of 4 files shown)


packages/webcomponents/src/components/checkout-wrapper/checkout-wrapper.tsx
```diff
@@ -44,44 +44,52 @@ export class CheckoutWrapper {
...
 
   @Method()
   async validate(): Promise<boolean> {
-    if (this.cardFormRef || this.bankAccountFormRef) {
-      const formToValidate = this.cardFormRef || this.bankAccountFormRef;
-      return formToValidate.validate();
-    } else {
-      throw new Error('CardForm component not found');
-    }
+    const validationResults = await Promise.all([
```
jakemer, line 58: Very clean! Nice.

packages/webcomponents/src/components/checkout-wrapper/checkout-wrapper.tsx
```diff
@@ -44,44 +44,52 @@ export class CheckoutWrapper {
...
+    if (!this.paymentMethodFormRef) {
+      return this.errorEvent.emit({
+        message: 'Payment method form not found',
         errorCode: ComponentErrorCodes.TOKENIZE_ERROR,
         severity: ComponentErrorSeverity.ERROR,
       });
     }
+
+    const isValid = await this.validate();
+    if (!isValid) throw new Error('Form is not valid');
```
jakemer, line 77: I think as long as client-side errors are shown upon validation, we should just return here and not throw an error. Especially since we have the event emitting already

packages/webcomponents/src/components/checkout-wrapper/checkout-wrapper.tsx
```diff
@@ -44,44 +44,52 @@ export class CheckoutWrapper {
...
+
+    const billingInfoValues = this.billingInformationFormRef
```
jakemer, line 79: Another way you could write this is `await this.billingInformationFormRef?.getValues() ?? {};` - the `??` is a "nullish coalescing operator" (just learned of it) which let's you write this line in a bit more concise of a way

jakemer (approved):
Very nice work!

packages/webcomponents/src/components/checkout-wrapper/checkout-wrapper.tsx
TiagoFuelber: Very good call! I'll Apply it.

PR #1045, opened 2025-05-20
Fix Analytics Cleanup error

colganmeanor:
### Fix Analytics Cleanup error

This PR prevents the analytics cleanup function from causing errors if it doesn't detect any event listeners.

Links
-----

Closes https://github.com/justifi-tech/engineering-artifacts/issues/236

Developer considerations
--------------

- On any task

Developer QA steps
--------------------

The best way to test this is to link local web components to the local dashboard using `pnpm link`

Build and link the web component library package.

- Run: `pnpm build && pnpm dev`.
- Then: `cd packages/webcomponents && pnpm link --global`.

Link dashboard - from the dashboard root directory run `pnpm link @justifi/webcomponents`

Open the browser dev console, and navigate to and from pages that contain web components, IE payments view, payouts view, terminals view, etc

- [x] Should no longer see the console error: `Uncaught TypeError: cannot read properties of undefined (reading 'forEach')`

### DIFF
--- a/packages/webcomponents/src/components/business-details/business-details.tsx
+++ b/packages/webcomponents/src/components/business-details/business-details.tsx
@@ -32,8 +32,8 @@ export class BusinessDetails {
   }
 
   disconnectedCallback() {
-    this.analytics.cleanup();
-  }
+    this.analytics && this.analytics.cleanup();
+  };
 
   private initializeGetBusiness() {
     if (!this.businessId || !this.authToken) {
--- a/packages/webcomponents/src/components/business-details/business-details.tsx
+++ b/packages/webcomponents/src/components/business-details/business-details.tsx
@@ -32,7 +32,7 @@ export class BusinessDetails {
   }
 
   disconnectedCallback() {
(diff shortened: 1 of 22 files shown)


packages/webcomponents/src/components/business-details/business-details.tsx
```diff
@@ -32,8 +32,8 @@ export class BusinessDetails {
   }
 
   disconnectedCallback() {
-    this.analytics.cleanup();
-  }
+    this.analytics && this.analytics.cleanup();
```
TiagoFuelber, line 35: Optional chain could be used here as well with the same effect.
e.g `this.analytics?.cleanup()`

TiagoFuelber (dismissed):
💯

packages/webcomponents/src/components/business-details/business-details.tsx
colganmeanor: Ah good idea! Updated!

TiagoFuelber (approved):
💯

PR #1078, opened 2025-07-04
Examples Refactor: (Task 1.2 / part 2) Add the remaining component example JSX templates

TiagoFuelber:
**Description**

This adds the remaining component example templates in JSX.

Links
-----
https://github.com/justifi-tech/engineering-artifacts/issues/577

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
I've run all the files myself to test it. `cd apps/examples && pnpm ts-node `src/examples/[file name]`.
But I think it's okay to skip QA on this because in the end we are going to run only one command for all.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -35,12 +35,6 @@ importers:
       nodemon:
         specifier: ^3.1.4
         version: 3.1.9
-      react:
-        specifier: ^18.2.0
-        version: 18.3.1
-      react-dom:
-        specifier: ^18.2.0
-        version: 18.3.1(react@18.3.1)
     devDependencies:
       '@types/express':
         specifier: ^4.17.17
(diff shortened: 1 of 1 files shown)


apps/component-examples/src/components/CheckoutsListComponent.tsx
```diff
@@ -0,0 +1,27 @@
...
+      <div class="column-preview">
+        <div class="list-component-wrapper">
+          <justifi-checkouts-list
+            account-id={subAccountId}
+            auth-token={webComponentToken}
+          ></justifi-checkouts-list>
+        </div>
+      </div>
+      <div class="column-output" id="output-pane">
+        <em>Checkouts list output will appear here...</em>
```
colganmeanor, line 23: I dont think we need to have the two-column-layout setup for the examples that render list components. 

If this is intentional, and this was done to show event usage in the list components, then maybe we just need a specific layout designed to show the list component in 75% of the screen and the output-pane in like 25% of the screen.

PR #1126, opened 2025-08-27
871 fix card formbank account form share tabid to the iframes to prevent cross tab iframe communication

TiagoFuelber:
Links
-----
https://github.com/justifi-tech/engineering-artifacts/issues/871

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------
Setup: `iframe-components` should be run locally from the `597-tokenize-payment-method-web-component-not-sending-all-values` branch.
This can be tested by opening a checkout in two different tabs. 

- [x] Tokenization should succeed.

**Backward compatibility**
Run `web-component-library` from the `main` branch, and run `iframe-components` from the `597-tokenize-payment-method-web-component-not-sending-all-values`
- [x] Checkout tokenization should work without errors.

### DIFF
--- a/packages/webcomponents/src/components/checkout/bank-account-form/bank-account-form.tsx
+++ b/packages/webcomponents/src/components/checkout/bank-account-form/bank-account-form.tsx
@@ -8,6 +8,7 @@ import { configState, waitForConfig } from "../../config-provider/config-state";
 export class BankAccountForm {
   @State() isReady: boolean = false;
   @State() iframeOrigin: string;
+  @State() tabId: string;
 
   private accountNumberIframeElement!: HTMLIframeInputElement;
   private routingNumberIframeElement!: HTMLIframeInputElement;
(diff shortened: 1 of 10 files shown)


packages/webcomponents/src/components/checkout/bank-account-form/bank-account-form.tsx
```diff
@@ -8,15 +8,17 @@ import { configState, waitForConfig } from "../../config-provider/config-state";
...
   @State() iframeOrigin: string;
+  @State() tabId: string;
 
   private accountNumberIframeElement!: HTMLIframeInputElement;
   private routingNumberIframeElement!: HTMLIframeInputElement;
 
   async componentWillLoad() {
     await waitForConfig();
     this.iframeOrigin = configState.iframeOrigin;
+    this.tabId = crypto.randomUUID?.() || `${Date.now()}-${Math.random().toString(36).slice(2)}`;
```
jakemer, line 19: Is the fallback in case of older browsers that do not have `crypto.randomUUID`? If so, should we create a util function to share the fallback between card and bank account forms? It probably doesn't matter as long as it's unique, but would still keep from repeating this string template

jakemer (dismissed):

PR #1223, opened 2025-12-02
Add new modular checkout layout examples and update documentation

TiagoFuelber:
- Introduced three new example components: ModularCheckoutLayout1Example, ModularCheckoutLayout2Example, and ModularCheckoutLayout3Example, showcasing different checkout layouts.
- Updated the index.ts file to export the new examples.
- Enhanced documentation to include live examples for each layout, improving usability and clarity for developers.
- Removed outdated example documentation to streamline content and focus on new implementations.

image

image

Links
-----

Developer considerations
--------------

- On any task
- On new features
  - [ ] Add unit tests
  - [ ] Add e2e tests
  - [ ] Add documentation
- On bugs

Developer QA steps
--------------------


packages/webcomponents/docs/helpers/ModularCheckoutLayout1Example.tsx
```diff
@@ -0,0 +1,478 @@
+import React, { useEffect, useRef, useState } from 'react';
```
jakemer: Should this be using react? That doesn't seem right since the web component library should not have react in it. The examples should be simple HTML and CSS that we can also show the code samples for, while keeping it easy to understand

PR #1253, opened 2025-12-12
fix: add loading state to payment provisioning form

jakemer:
## Summary
- Adds skeleton loading state while business data is being fetched in the payment provisioning form
- Prevents blank screen flash on initial component render
- Uses existing `PaymentProvisioningLoading` component that was already available but not utilized

## Changes
- Added `businessLoading` state to track business data fetch
- Updated render method to show skeleton loader while `businessLoading` is `true`
- Set `businessLoading` to `false` in both success and error callbacks

## QA Plan
- [ ] Navigate to payment provisioning component in Storybook
- [ ] Verify skeleton loading state displays immediately on load
- [ ] Verify form content appears after business data loads (no blank flash)
- [ ] Test with slow network to ensure loading state is visible
- [ ] Verify error cases still work correctly (loading state dismisses on error)

### DIFF
--- a/packages/webcomponents/package.json
+++ b/packages/webcomponents/package.json
@@ -39,7 +39,7 @@
   },
   "dependencies": {
     "@popperjs/core": "^2.11.8",
-    "@stencil/core": "^4.0.5",
+    "@stencil/core": "4.38.3",
     "@stencil/store": "^2.0.8",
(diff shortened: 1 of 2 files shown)


TiagoFuelber (approved):
💯

